PR #36, opened 2017-04-26
Bunch of updates for PushmiPullyu

murny:
TODO


spec/pushmi_pullyu_spec.rb
```diff
@@ -6,7 +6,7 @@
   end
 
   it 'does something useful' do
-    pending("the competion of some features")
```
cwant: Doh!

PR #40, opened 2017-04-26
Add rubocop to default rake task for travisCI

murny:
Also turned off Style/SymbolArray cop which forces array of symbols to use`%i` or `%I` instead of brackets


mbarnett:
:shipit:

PR #41, opened 2017-04-26
[DON'T MERGE] Initial spike of raw redis queue

mbarnett:
Using this PR as a staging area for the spike. It should not be merged at this point.

Here's what the raw redis queue version looks like:

Populating a Redis priority set (the Hydranorth `after_save` side of things, is just a call to `zadd`):
```ruby
 :001 > r = Redis.new
 => #<Redis client v3.3.3 for redis://127.0.0.1:6379/0> 
 :002 > 10.times do |i|
 :003 >     r.zadd 'pmpy_queue', Time.now.to_f, "noid#{i}"
 :004?>   end
 => 10 
 :005 > r.zrange('pmpy_queue', 0, -1)
 => ["noid0", "noid1", "noid2", "noid3", "noid4", "noid5", "noid6", "noid7", "noid8", "noid9"] 
```

The Pushmi-Pullyu side can be seen in this PR. A sample interaction with it looks like:

```ruby
 ./bin/console
 :001 > q = PushmiPullyu::PreservationQueue.new
 => #<PushmiPullyu::PreservationQueue:0x007fd2561c1a30 @redis=#<Redis client v3.3.3 for redis://127.0.0.1:6379/0>> 
 :002 > q.wait_next_item
 => "noid0" 
 :003 > q.wait_next_item
 => "noid1" 
# <etc>
# elements return immediately until the queue is empty, at which point
# wait_next_item blocks until there is something in the queue
```


lib/pushmi_pullyu/queue.rb
```diff
@@ -0,0 +1,53 @@
+require 'redis'
+
+# Quick and dirty take on sorted sets
+
+# 1) Create a sorted set in Redis (https://redis.io/topics/data-types). Call it preservation_queue
+#
+# 2) In GenericFile add an after_save that:
+#    - determines a monotonically increasing "score". Obvious scores would be either the time in seconds/milliseconds
```
murny: Just to clarify the score could just be a timestamp (in epoch time for example)?  Or would something more complicated be required for this?

mbarnett: It could be pretty arbitrary, as long as it increases (or decreases) over time to give some ordering to the queue.

I'm using epoch as float in this example, which gives us separation between events on the millisecond resolution. We could honestly just use the integer version, which would give us second resolution, with no real difference -- even if multiple events are added on the same second, this scheme still works because it doesn't really matter to us what order things come out of the queue at as long as things that get updated again "later" move to the back of the queue.

murny: Yeah agreed. Thansk for explanation, makes sense.

Don't think we need the granularity of `.to_f` gives. `.to_i` is probably more then good enough with the reasons you gave.

Issue #18, opened 2017-04-24
Gem PushmiPullyu with ability to run as daemon process


murny:
Gemifying of PushmiPullyu and running pushmipullyu as a CLI app with ability to run as a background process (daemon).

Maybe look into https://github.com/thuehlinger/daemons for helping with setting this up as a daemon.

(this will be pretty much done if we decided to go with Sidekiq implementation instead...see #17)

murny:
This is now done after final PRs have been merged. Closing issue.

PR #42, opened 2017-04-27
Specs for CLI class, bunch more refactors and adding daemon functionality

murny:
#18 

* Refactor CLI stuff and add some spec tests for it
* Extract default options into a config class
* Setup logger for PushmiPullyu
* Add daemon support using Daemons gem

### DIFF
--- a/exe/pushmi_pullyu
+++ b/exe/pushmi_pullyu
@@ -2,13 +2,6 @@
 
 require 'pushmi_pullyu'
 
-begin
-  cli = PushmiPullyu::CLI.instance
-  cli.parse
-  cli.run
-rescue => e
-  raise e if $DEBUG
-  STDERR.puts e.message
-  STDERR.puts e.backtrace.join("\n")
-  exit 1
-end
+cli = PushmiPullyu::CLI.instance
+cli.parse(ARGV)
+cli.run
--- a/exe/pushmi_pullyu
+++ b/exe/pushmi_pullyu
@@ -2,4 +2,13 @@
 
 require 'pushmi_pullyu'
 
-PushmiPullyu::CLI.new(ARGV).run!
+begin
+  cli = PushmiPullyu::CLI.instance
+  cli.parse
+  cli.run
+rescue => e
+  raise e if $DEBUG
+  STDERR.puts e.message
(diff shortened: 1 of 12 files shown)


exe/pushmi_pullyu
```diff
@@ -2,4 +2,13 @@
...
 require 'pushmi_pullyu'
 
-PushmiPullyu::CLI.new(ARGV).run!
+begin
+  cli = PushmiPullyu::CLI.instance
+  cli.parse
+  cli.run
+rescue => e
+  raise e if $DEBUG
+  STDERR.puts e.message
```
mbarnett: Are we just going to log to STDERR, or are we deferring logging to a later ticket?

lib/pushmi_pullyu/logging.rb
```diff
@@ -0,0 +1,33 @@
+require 'logger'
+
+# PushmiPullyu::Logging is a standard Ruby logger wrapper
+module PushmiPullyu::Logging
```
mbarnett: It occurred to me, if this is going to be a long-running daemon, we need to trap SIGHUP and swap out the logs to play nicely with log-rotation automation

exe/pushmi_pullyu
mbarnett: Nevermind, saw the logging class after. Are we intentionally bypassing it here?

murny: Yeah I'll get rid of this logic rescue logic. Not really needed.

lib/pushmi_pullyu/cli.rb
```diff
@@ -18,26 +18,47 @@ def initialize
...
+      exit(0)
     end
   end
 
   private
 
+  def setup_signal_traps
+    Signal.trap('INT') { raise Interrupt }
+    Signal.trap('TERM') { raise Interrupt }
+    Signal.trap('HUP') do
```
mbarnett: 👍

murny: I'll merge as is...but still figuring out how to reopen/rotate logs on SIGHUPs. So ill add another PR for this. As this one is getting pretty beefy as is

Issue #24, opened 2017-04-24
Read from Queue (Redis)


murny:
If we are custom implementing PushmiPullyu (see #17), We will need to implement the communication of PushmiPullyu to talk to our Queue (redis). 

PushmiPullyu will sit and monitor this queue. Once an element on the queue is "ready" (passes a certain window of wait period), PushmiPullyu will pop this element off the queue and start the preservation event.

murny:
Has been completed in PR #43. PushmiPullyu can now read/monitor the queue (redis) and retrieve items off it. Closing out issue.

Issue #17, opened 2017-04-24
SPIKE: Custom implementation vs using Sidekiq for PushmiPullyu App


murny:
If you look at the workflow documented here https://github.com/ualbertalib/pushmi_pullyu/issues/16, this is pretty much exactly how [Sidekiq](https://github.com/mperham/sidekiq) works. PushmiPullyu is essentially a background job processor.

- Jobs are enqueued in a redis queue (after save callback that pushes GF into redis queue)
- Sidekiq montiors and retrieves jobs from redis queue (what PushmiPullyu will do)
- Performs job (retrieve data from fedora, builds AIP, bags AIP, pushes AIP to Swift, and logs events)

Do we need to custom implement everything for PushmiPullyu? At very least we should be using Sidekiq and other job queues as ideas.

Pros for using Sidekiq: 

- Work for running PushmiPullyu on a server as a CLI and dameonizing it will not be required
- Proven and stable track record for running in production
- Fantastic Logs and built in Dashboard
- Easier to maintain and for new people to get trained on (most rails devs know how to use job queues)
- Sidekiq is thread based and concurrent (could be a concern too?)

Things to be concerned about regarding Sidekiq:

- Sidekiq does not support unique jobs out of the box, requires addons, etc. (pro version also has unique jobs)
  - https://github.com/mhenrixon/sidekiq-unique-jobs
- Time window period logic could be done when enqueuing the job from ERA (`PushmiPullyuJob.set().perform later()`)

Henry asks about replay support for failed jobs (sidekiq has this built in, how would we tackle this using a custom approach)?

mbarnett:
Here's what the raw redis version looks like

Populating a Redis priority set (the Hydranorth `after_save` side of things, is just a call to `zadd`:
```ruby
 :001 > r = Redis.new
 => #<Redis client v3.3.3 for redis://127.0.0.1:6379/0> 
 :002 > 10.times do |i|
 :003 >     r.zadd 'pmpy_queue', Time.now.to_f, "noid#{i}"
 :004?>   end
 => 10 
 :005 > r.zrange('pmpy_queue', 0, -1)
 => ["noid0", "noid1", "noid2", "noid3", "noid4", "noid5", "noid6", "noid7", "noid8", "noid9"] 
```

The Pushmi-Pullyu side can be seen in this PR. A sample interaction with it looks like:

```ruby
 ./bin/console
 :001 > q = PushmiPullyu::PreservationQueue.new
[...]

Issue #27, opened 2017-04-24
Tests and Specs (corner cases and VCR setup)


murny:
PushmiPullyu should come with automated tests and specs.

- Testing of queue and various corner cases (uniqueness, time windows, etc)
-  Tests around creation of AIP
-  Tests around bagging and tarring of AIP
- Swift/Fedora should be mocked using something like [VCR gem](https://github.com/vcr/vcr)
- Errors in workflow are being handled appropriately

murny:
This will be more of a review ticket once all the other tickets have been completed. 

Should make sure we are doing through testing on a lot of the fragile areas of the codebase. Also a few high level integration tests should be done as well?

Testing infrastructure has already been setup (mainly for #30 ). Most code has been pushed to master has already been tested decently well (sitting at > 80% code coverage). Chris added VCR gem for mocking fedora also.

PR #43, opened 2017-05-02
Preservation Queue

### LINKED ISSUE #14: SPIKE: Determine feasibility of using a priority queue in Redis to control preservation
Questions needing to be determined and answered.

Pros and Cons for team of why we should use Redis over a RDMS or something else?

How to do a priority queue and other things (unique set?) in Redis?

mbarnett:
This work should be ready for a real review now.

Closes #14, #17, #24


lib/pushmi_pullyu/cli.rb
```diff
@@ -123,11 +124,11 @@ def run_tick_loop
     @running = true # set to false by signal trap
 
     while @running
-      sleep(5)
-      logger.debug('.')
       # Preservation (TODO):
       # 1. Montior queue
       # 2. Pop off GenericFile element off queue that are ready to begin process preservation event
+      item = @queue.wait_next_item
```
murny: Your missing the initialization of @queue? Currently a nil object.

Can probably throw it in the initialize method?

```ruby
   attr_accessor :config
   attr_accessor :queue

  def initialize
    self.queue = PushmiPullyu::PreservationQueue.new # or pass in some options?
    self.config = PushmiPullyu::Config.new
  end
```

Then use the queue getter here:

```ruby
item = queue.wait_next_item
```
Correct?

lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -0,0 +1,79 @@
...
+class PushmiPullyu::PreservationQueue
+
+  class ConnectionError < StandardError; end
+
+  def initialize(connection: {}, pool_opts: { size: 1, timeout: 5 }, poll_interval: 10, age_at_least: 0,
+                 queue_name: 'dev:pmpy_queue')
+    # we use a connection pool even though we're not (currently) threading
+    # as it transparently provides for repairing connections if they are closed after long periods of inactivity
+    @redis = ConnectionPool.new(pool_opts) do
+      connection.reverse_merge!(host: 'localhost', port: 6379)
```
murny: We will have to add this to the PushmiPullyu::Config class and have these as the default redis config. But we should think about ways to override this? From command line options or a YAML file?

lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -0,0 +1,79 @@
...
+#
+# A sorted set will only ever contain a noid once, with whatever score it was last given. Because preservation_queue
+# is sorted lowest score to highest, and because scores increase over time, a cascade of jobs/updates will cause a noid
+# to keep "moving back" in the queue until it becomes the least recently updated noid in the queue, at which point it
+# will be popped and preserved. Any further updates will trigger a new AIP build.
+class PushmiPullyu::PreservationQueue
+
+  class ConnectionError < StandardError; end
+
+  def initialize(connection: {}, pool_opts: { size: 1, timeout: 5 }, poll_interval: 10, age_at_least: 0,
```
murny: We will have to add this to the PushmiPullyu::Config class and have these as the default redis config. But we should think about ways to override this? From command line options or a YAML file?

lib/pushmi_pullyu/cli.rb
mbarnett: This is what happens when you uncomment example code and call it a day.

lib/pushmi_pullyu/preservation_queue.rb
mbarnett: I was thinking we could feed them the command line for simplicity. Is there a separate ticket for getting all of the config stuff set up, or should I tackle that here?

murny: I'm currently parsing options off the command line which is updating the config class object (just using optparse as this was easiest instead of using something like Thor). So feel free to add to this.  This logic is all found in the cli.rb.

The config class (config.rb) will hold the defaults (and you can even add ENV VARS to override them there too). 

Shouldn't be much work to tackle this here? But let me know if we should create a ticket for this.

mbarnett: Nah I'll do it now

mbarnett: I didn't bother to add everything, but I added CLI option flags for the ones we're most likely to need to change immediately.

lib/pushmi_pullyu/cli.rb
```diff
@@ -120,14 +140,18 @@ def setup_log
   end
 
   def run_tick_loop
+    queue = PushmiPullyu::PreservationQueue.new(connection: { host: config.redis_host, port: config.redis_port },
```
murny: This queue initilize can move up into the start_server method. But very minor and we can do this later

spec/preservation_queue_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'spec_helper'
+require 'timecop'
+
+RSpec.describe PushmiPullyu::PreservationQueue do
```
murny: Should we do any unhappy path testing here? Or this can come later? Like in this ticket: https://github.com/ualbertalib/pushmi_pullyu/issues/27

lib/pushmi_pullyu/cli.rb
mbarnett: It could. I went back and forth, since it's never needed outside of this method so this scopes it tighter, but it would look cleaner in the init routines. Either way is good with me.

lib/pushmi_pullyu/cli.rb
```diff
@@ -89,6 +95,20 @@ def parse_options(argv)
         config.monitor = true
       end
 
+      opts.on('-rh', '--redis-host IP', 'Host IP of Redis instance to read from.'\
```
cwant: My only comment might be that the redis options could possibly moved to a separate method, then called here e.g. redis_options(opts)

murny: Not a bad idea. I think this is fine for now. But If this option parser continues to grow in size, we might need to do something to keep this from growing ridiculously large. 

Whether that is to just move it into a separate class/method. Or better yet, reserve the command line options for just dealing with running the program and any configuration with talking to redis/solr/fedora/swift etc to live in a config yaml file. Which it's location will be defaulted to `config/pushmi_pullyu.yml` or `config/pushmi_pullyu.yml.erb` and can be override on the command line:

```ruby
        opts.on '-C', '--config PATH', "path to YAML config file" do |config_file|
          config.config_file = config_file
        end
```
Then after the options parser we can do a override the options with the config file

This also gives us the ability to override for different environments (staging/production) and much more power. Instead of passing 20 different options on the command line to `pushmi_pullyu`

murny: Created issue for above commet here: https://github.com/ualbertalib/pushmi_pullyu/issues/47

cwant: I like what you have here. I guess in the example above, you would want the 'config' directory to be bundled as part of the gem? (If you don't do that, where does it get it's default config file location from?)

murny: Yeah exactly I'll have a default location with a default `config/pushmi_pullyu.yml`. If you ever need to change that then you can provide your own via command line. 

Aka `pushmi_pullyu -C config/production.yml`

PR #44, opened 2017-05-02
Add ability for rotating logs 

murny:
- Add logfile options for daemons, ability of reopening logfiles and SIGHUP now properly reopens logs
- Add graceful shutdown of run loop and do not catch interrupt exceptions
- Add specs for handling shutdown


lib/pushmi_pullyu/cli.rb
```diff
@@ -32,39 +32,37 @@ def run
 
   def start_server
     setup_signal_traps
+
     setup_log
     print_banner
 
-    begin
-      run_tick_loop
-    rescue Interrupt
```
murny: Feel like rescuing this exception is bad so handle this smarter....

Also allowing for a graceful exit instead of forcing it

lib/pushmi_pullyu/cli.rb
```diff
@@ -32,39 +32,37 @@ def run
...
-    Signal.trap('INT') { raise Interrupt }
-    Signal.trap('TERM') { raise Interrupt }
+    Signal.trap('INT') { Thread.new { shutdown } }
+    Signal.trap('TERM') { Thread.new { shutdown } }
     Signal.trap('HUP') do
-      if config.logfile
-        logger.debug 'Received SIGHUP, reopening log file'
-        # TODO: reopen logs
-        # PushmiPullyu::Logging.reopen_logs(config.logfile)
+      Thread.new do
```
murny: The Thread.new stuff is ulgy... but since Ruby 2+,  Mutex#lock cannot live within a signal handler trap context. This prevents Logger from working since it uses an underlying mutex.

More info here:
https://bugs.ruby-lang.org/issues/7917

So this is kinda of a hack.

Alternative solutions here:
https://gist.github.com/mvidner/bf12a0b3c662ca6a5784

Seen a few projects use the The self-pipe trick? Maybe we should do one of these solutions.

The shutdown ones can avoid using logger by using `$stderr.puts` or something. But the reopening of log is unavoidable.

lib/pushmi_pullyu/cli.rb
```diff
@@ -32,39 +32,37 @@ def run
...
+        PushmiPullyu::Logging.reopen
       end
     end
   end
 
+  def shutdown
```
murny: This allows for a graceful shutdown, by stopping the run loop and letting the program end itself. If you want to force shutdown, issue another INT/TERM to the program (CTRL-C or `pushmi_pullyu stop`, etc)

lib/pushmi_pullyu/cli.rb
```diff
@@ -140,11 +138,19 @@ def start_server_as_daemon
...
-                                          dir_mode: :normal,
-                                          monitor: config.monitor,
-                                          ARGV: @options) do |*_argv|
 
+    options = {
+      ARGV:       @parsed_opts,
+      dir:        config.piddir,
+      dir_mode:   :normal,
+      monitor:    config.monitor,
+      log_output: true,
```
murny: With these log options I am forcing everything from the daemon to the log file specified (default is /log/pushmi_pullyu.log). Previously exceptions/etc were not being logged? (or if they were going to somewhere like `/var/log`)

lib/pushmi_pullyu/cli.rb
```diff
@@ -32,39 +33,37 @@ def run
...
-    Signal.trap('INT') { raise Interrupt }
-    Signal.trap('TERM') { raise Interrupt }
-    Signal.trap('HUP') do
-      if config.logfile
-        logger.debug 'Received SIGHUP, reopening log file'
-        # TODO: reopen logs
-        # PushmiPullyu::Logging.reopen_logs(config.logfile)
-      end
-    end
+    Signal.trap('INT') { Thread.new { shutdown } }
```
murny: The Thread.new stuff is ulgy... but since Ruby 2+, Mutex#lock cannot live within a signal handler trap context. This prevents Logger from working since it uses an underlying mutex.

More info here:
https://bugs.ruby-lang.org/issues/7917

So this is kinda of a hack.

Alternative solutions here:
https://gist.github.com/mvidner/bf12a0b3c662ca6a5784

Seen a few projects use the The self-pipe trick? Maybe we should do one of these solutions.

The shutdown ones can avoid using logger by using $stderr.puts or something. But the reopening of log is unavoidable.

lib/pushmi_pullyu/cli.rb
```diff
@@ -32,39 +33,37 @@ def run
...
+    Signal.trap('TERM') { Thread.new { shutdown } }
+    Signal.trap('HUP') { Thread.new { PushmiPullyu::Logging.reopen } }
+  end
+
+  def shutdown
```
murny: This allows for a graceful shutdown, by stopping the run loop and letting the program end itself. If you want to force shutdown, issue another INT/TERM to the program (CTRL-C or pushmi_pullyu stop, etc)

lib/pushmi_pullyu/cli.rb
```diff
@@ -160,15 +130,54 @@ def run_tick_loop
...
+    else
+      logger.formatter = PushmiPullyu::Logging::SimpleFormatter.new
+    end
+    logger.level = ::Logger::DEBUG if config.debug
+  end
+
+  def setup_signal_traps
+    Signal.trap('INT') { Thread.new { shutdown } }
+    Signal.trap('TERM') { Thread.new { shutdown } }
+    Signal.trap('HUP') { Thread.new { PushmiPullyu::Logging.reopen } }
```
mbarnett: Complex question, but is this threadsafe? It's going to depend on the underlying logger, but if it's not, we could see some weirdness here. It looks like it could update @logger in one thread while another is writing through it, which could do some weird things (half of the logs from a single loop end up in one file, half in another, for one possibility). 

One way to deal with this might be to set a flag on the main run loop, and then reset the logger at the end or beginning of the main loop.

(Having flashbacks to CMPUT 379 right now)

lib/pushmi_pullyu/cli.rb
```diff
@@ -160,15 +130,54 @@ def run_tick_loop
...
+  end
+
+  def setup_queue
+    @queue = PushmiPullyu::PreservationQueue.new(connection: { host: config.redis_host, port: config.redis_port },
+                                                 queue_name: config.queue_name,
+                                                 age_at_least: config.minimum_age)
+  end
+
+  def shutdown
+    exit!(1) unless running?
```
mbarnett: Might be worth a comment to explain the logic here, it took me a sec to figure it out.

Also, `exit!(1)` signals an error to any supervising software. Is that the desired semantics on a "forced" shutdown? If it were being run by upstart or something, double-interrupting would force it to reboot?

lib/pushmi_pullyu/cli.rb
murny: Comments got squashed after merge: https://github.com/ualbertalib/pushmi_pullyu/pull/44#discussion_r114667160

But agree, we will do this smarter

lib/pushmi_pullyu/cli.rb
murny: Original PR comment got squashed here: https://github.com/ualbertalib/pushmi_pullyu/pull/44#discussion_r114667295

Yeah comment is worth having though!

Reason for exit!(1)? Since you force shutting it down I figure that's not a normal successful program termination. But we can change that to exit 0 if you think that's better

mbarnett: It almost depends on how they want to manage the daemon in prod. I don't think it's a big deal either way. Leave it this way for now and we can change it later if need be

lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -61,11 +61,11 @@ def next_item
   end
 
   def wait_next_item
-    loop do
-      element = next_item
-      return element if element.present?
-      sleep @poll_interval
-    end
+    # loop do
```
murny: Is a second loop needed @mbarnett ? Or can just using the main loop work fine here? Will remove comments if you agree if this is okay.

lib/pushmi_pullyu/logging.rb
```diff
@@ -29,8 +29,10 @@ def self.logger=(log)
 
   def self.reopen
     if @logger
+      logger.info 'reopen!'
```
murny: oops will remove these.. just some debugging

lib/pushmi_pullyu/preservation_queue.rb
mbarnett: The Redis client API doesn't block, so wait_next_item will return nil immediately if the queue is empty. I think it makes morse sense to poll in a loop inside that method for new items rather than handle the nil in the main loop?

murny: We need a way to close this loop down when a interrupt is given then? Some global flag that calls break if true or something?

mbarnett: Good point. That won't play well with the sleep/poll interval, either. Let me think about this for a bit

lib/pushmi_pullyu.rb
```diff
@@ -60,4 +60,24 @@ def self.logger
   def self.logger=(log)
     PushmiPullyu::Logging.logger = log
   end
+
+  def self.server_running=(boolean)
```
mbarnett: Maybe `status` or something instead of `boolean`? That's like naming a number `int` 🙃

lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -61,11 +61,11 @@ def next_item
   end
 
   def wait_next_item
-    # loop do
-    element = next_item
-    return element if element.present?
-    sleep @poll_interval
-    # end
+    while PushmiPullyu.continue_run_loop?
```
mbarnett: I know this name was my idea, but seeing it now maybe `continue_run_loop?` is the wrong name, considering that we don't use it in the main run loop and it can return false while we still want to continue the main run loop (when we need to rotate the logs).

`continue_polling?` maybe?

murny: Will change :+1: . Naming is hard

lib/pushmi_pullyu.rb
murny: haha fair enough, ill change

PR #46, opened 2017-05-03
Add redis to travicCI build to fix failing builds

murny:
Since we now use redis, Add redis-server service to travisCI to fix failing builds


murny:
CI is now passing!

Issue #2, opened 2017-02-23
Pull Data from Fedora/Solr


pbinkley:
_From @pbinkley on December 7, 2016 20:43_

- run a fixity report in the Fedora API; save the output. Stop if the datastream is already corrupt in Fedora.
- retrieve Fedora information (see spec document https://docs.google.com/document/d/15qVr8naWHvUI1wgTLn-53zo4QO9QH3GMjXDCeLGq_sQ/edit#heading=h.sqqxwzm2947u)

Validate all available checksums - e.g. for the main datastream, after it is retrieved.

[Bash script mockup](https://gist.github.com/pbinkley/060ed60fb580c8c26324d07048ea1518) of process of fetching content from Fedora and populating a local directory ready to be bagged

See #22 for AIP creation

_Copied from original issue: ualbertalib/HydraNorth#1330_

murny:
Closed by PR #52.

Issue #47, opened 2017-05-04
Handle PMPY configuration via YAML file


murny:
Add code to parse a YAML file and set options for multiple configurations we will need. Such as options (Host/port/url/etc) for Fedora/Swift/Redis/Solr.

Should maybe support the ability to for multiple environments? Production/Staging 

Issue raised in #43 PR (https://github.com/ualbertalib/pushmi_pullyu/pull/43#discussion_r114851407): 

"Not a bad idea. I think this is fine for now. But If this option parser continues to grow in size, we might need to do something to keep this from growing ridiculously large. 

Whether that is to just move it into a separate class/method. Or better yet, reserve the command line options for just dealing with running the program and any configuration with talking to redis/solr/fedora/swift etc to live in a config yaml file. Which it's location will be defaulted to `config/pushmi_pullyu.yml` or `config/pushmi_pullyu.yml.erb` and can be override on the command line:

```ruby
        opts.on '-C', '--config PATH', "path to YAML config file" do |config_file|
          config.config_file = config_file
        end
```
Then after the options parser we can grab the options from the config file

This also gives us the ability to override for different environments (staging/production) and much more power. Instead of passing 20 different options on the command line to `pushmi_pullyu`"

murny:
Moving this into close now that PRs have been merged.  Closing issue.

PR #48, opened 2017-05-04
[WIP -- don't merge] Issue #2 Fedora/Solr pull.

cwant:
Initial commit of Fedora config handling. PR-ing for feedback.

### DIFF
--- a/config/fedora.yml
+++ b/config/fedora.yml
@@ -1,0 +1,4 @@
+user: fedoraAdmin
+password: fedoraAdmin
+url: http://127.0.0.1:8983/fedora/rest
+base_path: /dev
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 require 'optparse'
 require 'singleton'
 
@@ -89,7 +90,8 @@ class PushmiPullyu::CLI
         config.monitor = true
       end
 
-      opts.separator ''
+      config.fedora.process_options(opts)
+
       opts.separator 'Common options:'
 
       opts.on_tail('-v', '--version', 'Show version') do
--- a/lib/pushmi_pullyu/fedora_config.rb
+++ b/lib/pushmi_pullyu/fedora_config.rb
@@ -1,0 +1,46 @@
+require 'yaml'
+class PushmiPullyu::FedoraConfig
+
+  include PushmiPullyu::Logging
+
+  attr_accessor :config_file, :url, :user, :password, :base_path
+
+  def initialize
+    self.config_file = ENV['FEDORA_CONFIG_FILE'] ||
+                       File.expand_path(File.dirname(__FILE__) + '../../../config/fedora.yml')
+
+    self.url = ENV['FEDORA_URL']
+    self.user = ENV['FEDORA_USER']
+    self.password = ENV['FEDORA_PASSWORD']
+    self.base_path = ENV['FEDORA_BASEPATH']
+  end
+
+  def process_options(opts)
+    opts.on('-fc', '--fedora-config-file FILE',
+            "Location of Fedora configuration (Default: #{config_file})") do |filename|
+      config_file = filename
+    end
+
+    opts.separator ''
+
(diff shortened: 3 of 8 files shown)


cwant:
I completely b0rked a rebase of this branch. Closing this PR and trying again ...

config/fedora.yml
```diff
@@ -0,0 +1,4 @@
+user: fedoraAdmin
```
murny: There's some duplication of work here. I created an issue for this config stuff here: #47 which I am currently working on.

Also I rather not have a yaml file for each piece of infrastructure we are using as that will be 4+ yaml files. (redis/solr/fedora/swift/etc). Why not just put everything in one yaml file?

lib/pushmi_pullyu/cli.rb
```diff
@@ -89,7 +90,8 @@ def parse_options(argv)
         config.monitor = true
       end
 
-      opts.separator ''
```
murny: Why remove this? Accident?

config/fedora.yml
mbarnett: I think ActiveFedora is hard-coded to load its own config file, unfortunately? That means we need at least a fedora.yml and a solr.yml and whatever the silly file names are for the test versions of those.

lib/pushmi_pullyu/fedora_config.rb
```diff
@@ -0,0 +1,46 @@
+require 'yaml'
+class PushmiPullyu::FedoraConfig
```
murny: Are you going to create a class for every config file too? Assuming a SolrConfig class is coming too? Might get a bit much no?

config/fedora.yml
murny: Guess it's to early to review this PR then since there's nothing about ActiveFedora here.

mbarnett: Uh nevermind, the ActiveFedora route would make this harder because it requires a 1-1 redeclaration of the original model

lib/pushmi_pullyu/fedora_config.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def process_options(opts)
+    opts.on('-fc', '--fedora-config-file FILE',
+            "Location of Fedora configuration (Default: #{config_file})") do |filename|
+      config_file = filename
+    end
+
+    opts.separator ''
+
+    load_from_config_file
```
murny: This should probably happen after the opt parser. Since you want `pushmi_pullyu --help` or `pushmi_pullyu --version` to do their thing and exit quickly. Now it will read from this file and parse all the opts, then execute the `--help/--version`

config/fedora.yml
cwant: Oops, these reviews seem to be for a PR that I closed first thing this morning. I will look at the comments though. Seems silly that Github would allow commenting on a closed PR. The current PR, which includes solr config is here: https://github.com/ualbertalib/pushmi_pullyu/pull/49

I'm indifferent as to whether there is one yaml or multiple (as long as the yaml tree doesn't get too branchy).

I thought we weren't going to use ActiveFedora (not sure there is any reason to), so the need to use separate YAML files may be moot.

Anyways, I need to read in some Fedora/Solr configuration somehow, so I'll continue down this road so that I'm not blocked by Shane's work on the config stuff (then I can switch over to his framework after he's done).

lib/pushmi_pullyu/cli.rb
cwant: Likely

lib/pushmi_pullyu/fedora_config.rb
cwant: Maybe. I personally prefer config.fedora.username to config.fedora_username as a conceptual, single-responsibility type of grouping. I had looked at a superclass for config loading to DRY it up a bit, but it wasn't going in a direction I liked (the subclasses still looked unnaturally complicated).

lib/pushmi_pullyu/fedora_config.rb
cwant: Right, my bad.

PR #50, opened 2017-05-08
#47: Handle PMPY configuration via yaml file

murny:
* Removed Config Class
* Added logic for handling and setting configuration via YAML
  * Adding 'pushmi_pullyu.yml` to config directory automatically loads
  * command line arguments take precedence over YAML configuration
* Refactor test suite and bring in rubocop-rspec


spec/pushmi_pullyu/cli_spec.rb
```diff
@@ -200,7 +200,7 @@
       end
 
       it 'results in an identical options hash, except for config_file' do
-        cli.parse
+        cli.parse([])
```
murny: Test was failing because it was using ARGV which in this case belongs to rspec and thus passing in rspec args: 
```
["--pattern", "spec/**{,/*/**}/*_spec.rb"]
```
Since this isn't a proper option for PushmiPullyu it fails. Instead let's just stub args here as an empty array.

examples/pushmi_pullyu.yml
```diff
@@ -0,0 +1,39 @@
...
+  host: localhost
+  port: 6379
+  queue_name: dev:pmpy_queue
+  minimum_age: 0
+
+solr:
+  url: http://localhost:8983/solr/hydra-development
+
+fedora:
+  url: http://localhost:8984/rest
```
cwant: Why 8984? It would seem more common to have fedora and solr served from the same port since they are often running in the same jetty/tomcat instance.

murny: These just example configs. Why it lives in examples folder. Think I stole them from hydranorth or something.

lib/pushmi_pullyu.rb
```diff
@@ -1,12 +1,63 @@
...
+      host: 'localhost',
+      port: 6379,
+      queue_name: 'dev:pmpy_queue'
+    },
+    # TODO: rest of these are examples for solr/fedora/swift... feel free to fill them in correctly
+    solr: {
+      url: 'http://localhost:8983/solr/hydra-development'
+    },
+    fedora: {
+      url: 'http://localhost:8984/rest',
```
cwant: (In regards to the other comment ...)
Both HydraNorth and Avalon have 8983 as Fedora port in their configs. It's no biggie, and can be overridden, but it would be nicer to have a default that works out of the box.

lib/pushmi_pullyu/cli.rb
```diff
@@ -9,19 +12,18 @@ class PushmiPullyu::CLI
...
-  end
+  def parse(argv = ARGV)
+    opts = parse_options(argv)
+    opts[:daemonize] = true if COMMANDS.include? argv[0]
+    opts = parse_config(opts[:config_file]).merge(opts) if opts[:config_file]
 
-  def parse(args = ARGV)
-    parse_options(args)
-    parse_commands(args)
+    options.merge!(opts)
```
cwant: I wonder if it makes sense to freeze the options at this point?

lib/pushmi_pullyu.rb
murny: Update them to proper defaults when you add your fedora/solr stuff. I literally just picked random defaults as none of this functionality exists yet. For example look at swift defaults haha.

cwant: Yeah, will do. (I now just noticed that the word 'fedora' isn't part of the path either.)

murny:
Yeah I prefer the dot notation myself. Reason why I went with the Config Class originally. But when loading from a yaml file (which gives you a hash), it's so much easier and less code to just work hashes with hashes. Don't need to do any crazy ruby metaprogramming (send/method missing/etc). Can just simply merge hashes together (and depending on which hash is merging into which hash, this makes determining precedence very easy too (CLI > YAML > Defaults).

examples/pushmi_pullyu.yml
```diff
@@ -0,0 +1,39 @@
...
+  user: fedoraAdmin
+  password: fedoraAdmin
+  base_path: /dev
+
+swift:
+  auth_version: 2.0
+  tenant: Millennium Falcon
+  username: han
+  password: YT-1300
```
mbarnett: Important question: did you steal this from somewhere or are you enough of a nerd to know the model number of the Millennium Falcon?

lib/pushmi_pullyu.rb
```diff
@@ -1,12 +1,63 @@
...
+    debug: false,
+    logfile: LOGFILE,
+    minimum_age: 0,
+    monitor: false,
+    piddir: PIDDIR,
+    process_name: PROCESS_NAME,
+    redis: {
+      host: 'localhost',
+      port: 6379,
+      queue_name: 'dev:pmpy_queue'
```
mbarnett: We don't have to add it in this PR, but we need to stick a note here and in my HN stuff that the default queue names should be kept in sync in both codebases.

lib/pushmi_pullyu.rb
```diff
@@ -1,12 +1,63 @@
...
+      temp_url_key: '492727ZED'
+    }
+  }.freeze
+
+  def self.options
+    @options ||= DEFAULTS.dup
+  end
+
+  def self.options=(opts)
+    @options = opts
```
mbarnett: Should this do a `.reverse_merge` with defaults so that you can set only some options and let the rest inherit the defaults? It would let you not handle the merging out in the parse command, which makes this a little cleaner.

lib/pushmi_pullyu.rb
mbarnett: Also, total bikeshed: the queue name is a concept independent of redis (we could replace the queue implementation with postgres or something and keep it), so to me it seems like it should be outside the `redis:` key, like the `minimum_age:` queue default.

spec/spec_helper.rb
```diff
@@ -9,4 +9,12 @@
   config.expect_with :rspec do |c|
     c.syntax = :expect
   end
+
+  config.disable_monkey_patching!
```
mbarnett: This method name is terrifying compared to what the docs says it actually does...

lib/pushmi_pullyu/cli.rb
mbarnett: This is the merge call that could be encapsulated inside the `options=`, I think?

examples/pushmi_pullyu.yml
murny: I stole it :sweat_smile: 

https://github.com/memoways/swift-ruby#usage

lib/pushmi_pullyu.rb
murny: Yeah wasn't sure about that. I felt minimum_age is more of a program option then redis, so made it top level and queue was more redis. But totally can change so both are top level options!

spec/spec_helper.rb
murny: Haha yeah not sure why they named it that. Basically just forces you to use `Rspec.describe` vs `describe` on the top level spec definition. Aka prevents exposing Rspec DSL globally

lib/pushmi_pullyu/cli.rb
murny: Oooh thats not a bad idea? Kinda like that.

lib/pushmi_pullyu.rb
murny: Yeah agree. I'll add this to PR.

murny: So i tried to implement this. My tests are breaking as I am using this setter to reset the state of `PushmiPullyu.options` after messing with them in my specs.

Mmm is this really that worth it? basically just abstracts the merge?  ¯\_(ツ)_/¯.

Ima pull this in without this feature. Can come back to it later if still worth it (add a new method maybe).

lib/pushmi_pullyu/cli.rb
murny: This means parse method can never be called more then once. Not sure if that is desired behaviour.

Plus freezing objects/hash are kinda weird in ruby. It's a frozen reference. Which mean's you can still assign a new hash to replace the original hash. It's not immutable like how you would think.

I'm going to leave it as is?

lib/pushmi_pullyu.rb
mbarnett: Isn't that a code smell? Now client code needs to set everything, and your tests need to reach into this and know how to reset options by calling dup on exposed variables. 

It looks to me like this should be pushed into a tiny utility class which encapsulates the defaults, handles the merging with defaults, and provides a method to *reset* the defaults, without pushing that code out to the clients.

spec/pushmi_pullyu/cli_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+      expect(cli).to have_received(:run_tick_loop).once
+    end
+  end
+
+  describe '#parse' do
+    let!(:opts) { PushmiPullyu.options.dup }
+
+    after do
+      PushmiPullyu.options = opts
+    end
```
mbarnett: This is what looks fishy to me. We're copying data out of an object and then writing it back in afterwards in a way that is really not demeter-friendly. We should encapsulate this interaction:

`PushmiPullyu.reset_options`

murny: Agreed will fix this and options setter!

PR #52, opened 2017-05-09
Fedora/Solr fetching, AIP bagging/tarring

cwant:
The main entry point is the create method in  PushmiPullyu::AIP, e.g.,

```ruby
# Download, bag, and tar the AIP and cleanup work directories
aip_filename = PushmiPullyu::AIP.create("v118rd525whatever")

# You can also get the AIP filename with ...
PushmiPullyu::AIP.aip_filename("v118rd525whatever")

# The AIP filename will typically be "tmp/work/v118rd525whatever.tar"

# Or if we want to create it but keep the work directories
aip_filename = PushmiPullyu::AIP.create("v118rd525whatever", should_clean_work_directories: false)

# The work files will be in ...
PushmiPullyu::AIP.aip_directory("v118rd525whatever")

# The AIP directory will typically be "tmp/work/v118rd525whatever"

# Cleanup everything (files, directories) ...
PushmiPullyu::AIP.destroy("v118rd525whatever")
```

See classes PushmiPullyu::AIP::Creator and PushmiPullyu::AIP::Download for the details.


lib/pushmi_pullyu/solr_fetcher.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    raise PushmiPullyu::SolrFetchError
+  end
+
+  # Return output as an array of hashes
+  def fetch_query_array(query, fields: nil, url_extra: nil)
+    out = run_query(query, fields: fields, url_extra: url_extra, csv: true)
+    results = []
+    header = nil
+    CSV.parse(out) do |row|
```
mbarnett: Regardless of CSV vs JSON, an explanatory comment block explaining the expected structure of the response data, so that the reader can make sense of the data munging below, would be great.

pushmi_pullyu.gemspec
```diff
@@ -27,6 +27,9 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'daemons', '~> 1.2', '>= 1.2.4'
   spec.add_runtime_dependency 'redis', '~> 3.3'
   spec.add_runtime_dependency 'connection_pool', '~> 2.2'
+  spec.add_runtime_dependency 'rdf-n3', '~> 1.1'
+  spec.add_runtime_dependency 'bagit', '~> 0.4'
+  spec.add_runtime_dependency 'minitar', '~> 0.6'
```
mbarnett: Let's make sure these get sorted by dependency name.

spec/spec_helper.rb
```diff
@@ -2,6 +2,9 @@
 require 'pushmi_pullyu'
 require 'pry'
 
+# Require support scripts
+Dir[File.dirname(__FILE__) + '/support/**/*.rb'].each { |f| require f }
```
mbarnett: Am I crazy or am I not seeing any `spec/support` directory

lib/pushmi_pullyu/aip_downloader.rb
```diff
@@ -0,0 +1,147 @@
...
+
+class PushmiPullyu::AipDownloader
+
+  include PushmiPullyu::AipDownloaderFilesDirectories
+
+  attr_reader :noid, :config, :logger, :aip_logger, :fetcher,
+              :aipcreation_log
+
+  attr_accessor :got_characterization, :got_thumbnail,
+                :got_fedora3foxml, :got_fedora3foxml_metadata
```
cwant: The ability to fetch these is optional. If we can fetch them, we will archive them, if not we don't raise an error. These track which ones of the optional pieces were successfully fetched. That said, they aren't used anywhere else in the code, so may not be useful.

lib/pushmi_pullyu/aip_downloader.rb
```diff
@@ -0,0 +1,147 @@
...
+
+    @config = config || PushmiPullyu.options
+    @logger = application_logger || PushmiPullyu.logger
+    @fetcher = PushmiPullyu::FedoraObjectFetcher.new(self.noid, self.config)
+
+    initialize_directory_names
+    initialize_filenames
+
+    # We need to log to application log and to the AIP log
+    @aip_logger = PushmiPullyu::AipLogger.new(@noid, aipcreation_log, @logger)
```
cwant: This logger is short lived and tied to the object that is being created. After the downloads, the log that is produced is part of the AIP. This is why I think the second logger is a good idea.

lib/pushmi_pullyu/aip_downloader.rb
```diff
@@ -0,0 +1,147 @@
...
+    fetcher ||= @fetcher
+    aip_logger.log_fetching(path)
+    fetcher.download_object(download_path: path,
+                            url_extra: url_extra,
+                            rdf: rdf)
+    aip_logger.log_saved(path)
+  end
+
+  def fetch_and_log_optional(path, url_extra: nil, fetcher: nil, rdf: false)
+    fetcher ||= @fetcher
```
cwant: The instance variable fetches for the current object. When we figure out what the permission objects are for the current object, we need a separate fetcher to get data from them to include in the AIP of the current object (and logged in the log for the current AIP). This is why there is an option to use a different fetcher.

lib/pushmi_pullyu/aip_downloader_files_directories.rb
```diff
@@ -0,0 +1,91 @@
...
+  CHARACTERIZATION_PATH = '/characterization'.freeze
+  VERSIONS_PATH = '/content/fcr:versions'.freeze
+  CONTENT_PATH = '/content'.freeze
+  THUMBNAIL_PATH = '/thumbnail'.freeze
+  FEDORA3FOXML_PATH = '/fedora3foxml'.freeze
+  FEDORA3FOXML_METADATA_PATH = '/fedora3foxml/fcr:metadata'.freeze
+
+  FILENAME_PREDICATE = 'info:fedora/fedora-system:def/model#downloadFilename'.freeze
+
+  def self.included(_klass)
```
cwant: The class was getting super-long and the logic was getting obscured by a lot of setup work, so I decided to spin this off as a module (not for reuse, but to keep semantically similar code/vars/constants together)

lib/pushmi_pullyu/aip_downloader_files_directories.rb
```diff
@@ -0,0 +1,91 @@
...
+                :tar_filename, :main_object_filename, :fixity_report_filename,
+                :content_datastream_metadata_filename,
+                :characterization_filename, :versions_filename,
+                :thumbnail_filename, :fedora3foxml_filename,
+                :fedora3foxml_metadata_filename
+  end
+
+  def make_object_directories
+    clean_directories
+    logger.debug("#{noid}: Creating directories ...")
```
cwant: Yeah, this is the application logger (not the logger that will produce a creation log for the current AIP). The AIP creation log can't be used at this point, since the directory it will live in doesn't exist yet. (Maybe these logging statements aren't needed, they are set to 'debug')

spec/spec_helper.rb
cwant: Not sure what you mean, spec/support/vcr.rb is in there.

lib/pushmi_pullyu/solr_fetcher.rb
cwant: Meh, just doing what Peter's example script does (it asked for CSV, so I'm asking for CSV).

lib/pushmi_pullyu/fedora_object_fetcher.rb
```diff
@@ -0,0 +1,59 @@
...
+      file.write(response.body)
+      file.close
+    else
+      puts response.body
+    end
+
+    true
+  end
+
+end
```
cwant: In this case we fetch from separate objects (main object, permission objects), so different instantiations are used.

lib/pushmi_pullyu/cli.rb
```diff
@@ -104,6 +104,10 @@ def parse_options(argv)
         opts[:piddir] = piddir
       end
 
+      o.on('-W', '--workdir PATH', 'Path where downloads, etc. are done ') do |workdir|
+        opts[:workdir] = workdir
```
cwant: I modelled this after what you did for the config loading of piddir ...

lib/pushmi_pullyu/aip_downloader_files_directories.rb
mbarnett: It's more confusing this way, honestly.

lib/pushmi_pullyu/aip_downloader.rb
mbarnett: Ok, I went and hunted down all of the callsites to understand this indirect approach to fetching. It looks like only one (a `fetch_and_log` call in `download_permission `) actually uses the fetcher parameter.

Is this indirection worth the confusion it causes, for that one call? My take right now is that this should be rewritten to be less confusing, because it reminds me a lot of the tangle of indirection in Hydra/Sufia, and is gaining you almost nothing.

lib/pushmi_pullyu/aip_downloader.rb
mbarnett: I'm not sold on the idea of a "short lived logger", especially with an eye towards potentially wanting to multithread this down the road. Synchronizing locks across a bunch of short lived "logger instance" foam is a lot more complicated, which is why logging is usually built as a cross-cutting, application-wide concern. 

Plus, two completely different logging definitions in different files is more-or-less the opposite of Single Points of Truth and really complicates the job of understanding this codebase for future readers.

spec/spec_helper.rb
mbarnett: I mean I'm blind/crazy

lib/pushmi_pullyu/solr_fetcher.rb
mbarnett: Well, now's as good an opportunity to consider if there's a better approach as any

mbarnett: *extremely your Mom's voice*: If Peter threw an AIP off a bridge, would you?!

lib/pushmi_pullyu/aip_downloader.rb
cwant: It's logging both to the application log and to the AIP specific log. Maybe the hangup is that I spun the thing off to a class with "Logger" in it's name. All it's doing is writing log statements to both the application log and to the AIP creation log. I'm not sure about the multithread comment. While it's true that multiple threads could write to the same application logger, I don't think two threads would write to the AIP creation log.

lib/pushmi_pullyu/fedora_object_fetcher.rb
```diff
@@ -0,0 +1,59 @@
+require 'pushmi_pullyu'
+require 'net/http'
+require 'fileutils'
+
+class PushmiPullyu::FedoraObjectFetcher
```
cwant: This is a really good idea.

lib/pushmi_pullyu/aip_downloader.rb
mbarnett: My hangup is that we're approaching logging in different ways in different places for no good reason. There's not even consistency with regards to formatting between the two logging approaches.

> While it's true that multiple threads could write to the same application logger, I don't think two threads would write to the AIP creation log.

How so? If two items are being preserved at roughly the same time, they would both be writing to the log in an indeterminate order. You're not creating a log per-object are you? The planned  approach was that we would log preservation actions to a single common preservation log, separate from the general application log, with an eye towards moving the preservation log elsewhere at a later date, and the Librarians grepping the common log for a noid of interest for the time being.

lib/pushmi_pullyu/fedora_object_fetcher.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    if download_path
+      file = File.open(download_path, 'wb')
+      file.write(response.body)
+      file.close
+    else
+      puts response.body
+    end
+
+    true
```
cwant: The downloaded object gets put in the 'download_path' option (or to stdout if that option isn't supplied). The boolean says whether it downloaded something. This only matters for optional files for the AIP (that is, when this is called with the return_false_on_404 option -- if that option isn't supplied, the method raises an error on 404).
Considering that the downloaded thing might be huge, there is an argument against returning it as an object from this method (particularly since we don't really do anything with the downloaded content except store it to disc).

lib/pushmi_pullyu/cli.rb
murny: This is fine. In case we ever want to override this from CLI

lib/pushmi_pullyu/fedora_object_fetcher.rb
mbarnett: Yeah, that make sense, but the name of the method is a bit misleading. `download_to_path` maybe? It also might be helpful to use a did_succeed type flag variable to hold the status, to make it a bit clearer that that's what this method returns. An explanatory comment on the method wouldn't hurt, either.

lib/pushmi_pullyu/fedora_object_fetcher.rb
```diff
@@ -0,0 +1,59 @@
+require 'pushmi_pullyu'
```
murny: You are requiring pushmi_pullyu on every one of your files in lib. This probably shouldn't be needed?

lib/pushmi_pullyu/aip_downloader.rb
cwant: Partially wrong. Yes, I am writing to "the single log", but each AIP also has it's own log that then gets added to the AIP and checksummed/manifested. See Peter's script and also the comments in slack about 'aipcreation log'. This model is trying to prevent having to do two logging statements all the time while the AIP is being assembled.

lib/pushmi_pullyu/cli.rb
mbarnett: Derp, right. Maybe we should rename that whole thwack of constants. DEFAULT_PIDDIR, DEFAULT_WORKDIR, etc? It's easy to come away with the impression that those are constants for the thing instead of constants for the defaults of those things.

lib/pushmi_pullyu.rb
```diff
@@ -5,11 +5,13 @@
 require 'pushmi_pullyu/version'
 require 'active_support'
 require 'active_support/core_ext'
+require 'pushmi_pullyu/aip_creator'
```
murny: Yeah i agree. Probably should be sorted in groups like mixins/modules (logging), class lib stuff, external library stuff (active support). All sorted alphabetically (A to Z)

lib/pushmi_pullyu/aip_downloader.rb
mbarnett: That makes sense as far as it goes, but this still needs to be folded into the main logging file. Nothing about that is intended to mean you have to "make two logging statements", there's no reason you can't push the "write two streams" logic into the one logger. Add methods and logic to write to multiple files and anything else you need, but do it in the one place that's related to logging in the codebase, rather than adding a second one.

Think about it from the perspective of a new reader to the codebase: you see a log call, hit cmd-t and start typing "logger" to find the relevant file, and then you discover that there's *multiple* loggers and you get to dig through 2, 3, 7, however many definitions of different loggers, done in different styles, to figure out which one is this one. What do we gain from pushing that confusion on future readers? What do we gain from establishing a pattern of creating different logging classes for every need? Let's keep this codebase as clean and simple as we can make it.

lib/pushmi_pullyu/cli.rb
murny: Yeah they probably not needed. They can just get folded into the default hash honestly

lib/pushmi_pullyu/aip_downloader.rb
cwant: A similar comment can be made about adding this stuff to the main application logger: you see a log call and you don't know which files it is writing to! (Which would also be hard to trace.) One of the log files it was writing to may not even exist a minute later. I'm not saying that the current state of this code can't be reworked, I just think you are going down the wrong path by requesting that the main application logger write to two streams (one permanent, the other temporary). At least when you log to "something else", this volatility is better recognized.

lib/pushmi_pullyu/aip_creator.rb
```diff
@@ -0,0 +1,73 @@
...
+class PushmiPullyu::AipCreator
+
+  attr_reader :noid, :config, :logger,
+              :workdir, :basedir, :tar_filename
+
+  def initialize(noid, config = nil, logger = nil)
+    @noid = noid
+
+    @config = config || PushmiPullyu.options
+    @logger = logger || PushmiPullyu.logger
```
cwant: I guess part of it is to make testing easier. Since we don't have production/development/test environment switching, this seemed like a good way to go to test individual classes/objects in isolation. See for example, overriding the config here: https://github.com/ualbertalib/pushmi_pullyu/blob/cwant/issue_2_pull_fedora_solr3/spec/pushmi_pullyu/fedora_object_fetcher_spec.rb#L4

lib/pushmi_pullyu/aip_downloader.rb
mbarnett: We might be arguing at cross-purposes here. When I say "I don't know what logger it's writing to" in another comment, I mean, it's unclear because we're tossing different loggers around through these instance variables and method parameters, sometimes shadowing the global logger through lazily initiated cache variables, so it's not clear at that call site what logger is an instance of, and the fact that it's a method param implies that it's unstable between calls.

What I'm saying would be clearer would be a situation in which all logging goes through the singleton `PusmiPullyu.logger`, since that was the entire motivation for creating the singleton in the first place. If you call `PusmiPullyu.logger.log_preservation_action('blah blah blah')` here, while it's true that I may not know which file(s) that call writes to, I would at least be clear on where to look without inspecting this method parameter at runtime to try to sort out what's happen where.

lib/pushmi_pullyu/fedora_object_fetcher.rb
murny: Yeah good suggestion. I think a namespace for PushmiPullyu::AIP:: would be a good addition for this. This helps organise all this code into a AIP folder (and you can remove all the `aip_` names too)

lib/pushmi_pullyu/aip_logger.rb
```diff
@@ -0,0 +1,49 @@
+require 'rdf'
```
murny: RDF or RDF/N3 stuff is never used in this class?

lib/pushmi_pullyu/aip_logger.rb
```diff
@@ -0,0 +1,49 @@
...
+require 'rdf/n3'
+require 'fileutils'
+require 'pushmi_pullyu/fedora_object_fetcher'
+require 'pushmi_pullyu/solr_fetcher'
```
murny: All these other ones arent used either.

lib/pushmi_pullyu/aip_logger.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class PushmiPullyu::AipLogger
```
murny: Yeah I would echo this statement too. Not seeing much benefit for this? Alot of this stuff can be added to the main logging.rb class if required. Preservation event as well.

lib/pushmi_pullyu/aip_downloader_files_directories.rb
```diff
@@ -0,0 +1,91 @@
+require 'rdf'
```
murny: We are not including this in our gemspec as a dependency if we are using it? 

Everything relating to RDF should be done via RDF/N3 gems interface? `RDF::N3`? Else we should be depending on this ourselves (incase RDF::N3 does something weird in the future) and adding RDF to our gemspec.

lib/pushmi_pullyu/aip_downloader_files_directories.rb
```diff
@@ -0,0 +1,91 @@
...
+    FileUtils.mkdir_p(metadatadir)
+    FileUtils.mkdir_p(logsdir)
+    FileUtils.mkdir_p(thumbnailsdir)
+    logger.debug("#{noid}: Creating directories done")
+  end
+
+  def clean_directories
+    return unless File.exist?(basedir)
+    logger.debug("#{noid}: Nuking directories ...")
+    FileUtils.rm_rf(basedir)
```
murny: I agree with a lot of this feedback. mixin module is leaving me very confused. (especially when its only being mixed/included in one place). Makes code very hard to trace and follow. (things are being instantized in a different class etc., things are just being called, etc.). If the main class is getting to big, maybe extract some of this directory functionality into a directory class for cleaning/making directories? Or look at some of matt's feedback.

lib/pushmi_pullyu/fedora_object_fetcher.rb
```diff
@@ -0,0 +1,59 @@
...
+
+class PushmiPullyu::FedoraObjectFetcher
+
+  attr_accessor :config, :noid
+
+  RDF_FORMAT = 'text/rdf+n3'.freeze
+
+  def initialize(noid, config = nil)
+    self.noid = noid
+    self.config = config || PushmiPullyu.options
```
murny: I would keep the same name of "options" too instead of using "config"? 

It's better for a CTRL-F and seeing whos using stuff....for example if i wanted to refactor `options[:fedora][:url]` to say....`optons[:fedora][:host]:.optons[:fedora][:port]` etc. Much harder if there named differently etc.

lib/pushmi_pullyu/aip_creator.rb
mbarnett: It makes understanding the non-test code path (ie, what actually happens in production) much more complicated, though. When every class and method get a logger argument passed into it, and has logger instance variables stored in it, you create a bunch of uncertainty as to what the actual object instance receiving all of these calls is. It's also inconsistent with the style being used elsewhere in this codebase.

The singletons offer an easily mockable interface for testing purposes. Adding runtime indirection to production just complicates the abstraction. significantly. Ditch it.

lib/pushmi_pullyu/aip_downloader.rb
mbarnett: YAGNI unless proven otherwise, in my opinion

lib/pushmi_pullyu/fedora_object_fetcher.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+      raise PushmiPullyu::FedoraFetchError
+    end
+
+    if download_path
+      file = File.open(download_path, 'wb')
+      file.write(response.body)
+      file.close
+    else
+      puts response.body
```
murny: Should be passed into logger instead of STDOUT?

lib/pushmi_pullyu/aip_downloader.rb
```diff
@@ -0,0 +1,147 @@
...
+  def fetch_and_log(path, url_extra: nil, fetcher: nil, rdf: false)
+    fetcher ||= @fetcher
+    aip_logger.log_fetching(path)
+    fetcher.download_object(download_path: path,
+                            url_extra: url_extra,
+                            rdf: rdf)
+    aip_logger.log_saved(path)
+  end
+
+  def fetch_and_log_optional(path, url_extra: nil, fetcher: nil, rdf: false)
```
murny: What's difference between `fetch_and_log` and `fetch_and_log_optional`? Looks like one is for mandatory fields for the AIP, other are optional ones (thumbnails). They look almost the exact same except you are passing `true` in for `return false on 404` and calling a different logging method. Could these be combined somehow?

lib/pushmi_pullyu/solr_fetcher.rb
```diff
@@ -0,0 +1,55 @@
+require 'pushmi_pullyu'
+require 'net/http'
+require 'csv'
+
+class PushmiPullyu::SolrFetcher
```
murny: Maybe keep fedora/solr file names more similar? SolrObjectFetcher? or FedoraFetcher?

lib/pushmi_pullyu/aip_creator.rb
```diff
@@ -0,0 +1,73 @@
...
+    destroy_aip_file
+  end
+
+  def destroy_aip_file
+    FileUtils.rm(tar_filename) if File.exist?(tar_filename)
+  end
+
+  def destroy_aip_directory
+    aip_downloader.clean_directories
+  end
```
murny: Maybe have this return the path to the tar file (call `tar_filename`/`aip_filename` at the very end)?

Then when a someone is accessing your stuff in CLI.rb or whereever it will just be:
```
aip_file = PushmiPullyu::AipCreator.new(noid, options, logger).create_aip
#do swift stuff => SwiftUpload.new(aip_filename)?
```

murny: I agree with this.  

You could enhance `create_aip` even to have it return the path to the tar file (call `tar_filename`/`aip_filename` at the very end of that method)? 

Then when a someone is accessing your stuff in CLI.rb `run_tick_loop` method or wherever it will just be:
```
aip_tar_file = PushmiPullyu::AipCreator.new(noid, options, logger).create_aip
#do swift stuff => SwiftUpload.new(aip_tar_file)?
```

lib/pushmi_pullyu/logging.rb
```diff
@@ -23,6 +23,18 @@ def self.logger
     @logger ||= initialize_logger
   end
 
+  def self.log_aip_activity(noid, message)
```
murny: Looks much better! Nice work :+1:

lib/pushmi_pullyu/aip.rb
```diff
@@ -0,0 +1,66 @@
...
+require 'archive/tar/minitar'
+
+module PushmiPullyu::AIP
+  # Exceptions
+  class BagInvalid < StandardError; end
+  class NoContentFilename < StandardError; end
+  class FedoraFetchError < StandardError; end
+  class SolrFetchError < StandardError; end
+
+  def self.create(noid, skip_download: false, clean_work_directories: true)
```
mbarnett: Minor tweak: give `skip_download` and `clean_work_directories` predicate names, like `should_skip_download`  or something

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -0,0 +1,214 @@
...
+    ).freeze
+  end
+
+  def content_filename
+    return @content_filename unless @content_filename.nil?
+
+    # Extract filename from main object metadata
+    graph = RDF::Graph.load(full_local_path(aip_paths.main_object.local))
+    graph.each_statement do |statement|
+      if statement.predicate == FILENAME_PREDICATE
```
mbarnett: Would it be better/faster to try using [`RDF::Graph#query`](https://github.com/ruby-rdf/rdf/blob/develop/examples/query_bench.rb#L13), or is iteration better? I haven't tried it but it seems like iteration could be slow on a large RDF graph.

lib/pushmi_pullyu/aip.rb
```diff
@@ -0,0 +1,40 @@
...
+require 'archive/tar/minitar'
+
+module PushmiPullyu::AIP
+  # Exceptions
+  class BagInvalid < StandardError; end
+  class NoContentFilename < StandardError; end
+  class FedoraFetchError < StandardError; end
+  class SolrFetchError < StandardError; end
+
+  def self.create(noid, should_skip_download: false, should_clean_work_directories: true)
```
mbarnett: I'm being really paranoid, but can we add a bit of validation here to verify that `noid` isn't empty? The queue should never yield an empty string, but if it did, the logic of `aip_directory` will lead to the wrong directory being `rm -rf`'d. This extends to a few other places in the codebase where an empty string and a badly chosen or missing `work_dir` could lead to accidentally deleting the wrong directory

cwant: Really good point. I should also do some validation so we don't noids like '../../etc/password'.
Is there a validator for what noid content can be? I'm guessing ^[A..Za..z0..9]+$ ... don't know if there are any other allowed characters.

lib/pushmi_pullyu/aip/creator.rb
```diff
@@ -0,0 +1,69 @@
...
+require 'pushmi_pullyu/aip'
+
+class PushmiPullyu::AIP::Creator
+
+  attr_reader :aip_filename, :aip_directory
+
+  def initialize(noid)
+    @noid = noid
+    @aip_directory = "#{workdir}/#{@noid}"
+    @aip_filename = "#{workdir}/#{@noid}.tar"
```
murny: Couldn't you access these from the `PushmiPullyu::AIP` module?

Like `PushmiPullyu::AIP.aip_filename` etc? Or are these different? Seems to be some duplication

cwant: Absolutely, should be DRY-er.

lib/pushmi_pullyu/aip.rb
```diff
@@ -0,0 +1,40 @@
...
+    # Return the filename of the created AIP tarball
+    PushmiPullyu::AIP::Creator
+      .new(noid)
+      .run(should_skip_download: should_skip_download,
+           should_clean_work_directories: should_clean_work_directories)
+    aip_filename(noid)
```
murny: Your `PushmiPullyu::AIP::Creator.new(...).run(....)` returns a `aip_filename` Why not just use this one? Or is this `aip_filename(noid)` different some how?

lib/pushmi_pullyu/aip/downloader.rb
cwant: Not sure. I would think that #query would want to load the entire graph to handle arbitrary queries, whereas iteration can break as soon as it hits the statement it wants (I don't know enough about this library to say for sure).

I also tried running #query to match what is happening in this method, but I am either doing it wrong or it was designed for a different use case.  E.g., refactoring the code to use:

```ruby
graph.query(predicate: FILENAME_PREDICATE)
```
returns an empty RDF::Queryable::Enumerator when executed via the test suite.

lib/pushmi_pullyu/aip/solr_fetcher.rb
```diff
@@ -0,0 +1,48 @@
+require 'pushmi_pullyu'
+require 'net/http'
+require 'json'
+
+class PushmiPullyu::AIP::SolrFetcher
+
+  def self.get_permission_object_ids(noid)
```
murny: Why couldn't you replace these self methods (which proxy to a new instance of the class and then call the same method but just on the instance) to do something like you did in the `fedora_fetcher.rb`? 

For example:

```
PushmiPullyu::AIP::SolrFetcher.new(noid).get_permission_object_ids

# calls simple intializer which assigned noid to instance variable 
# therefore you don't need to pass it in as an argument for `get_permissions_object_ids` instance method
```
vs

```
PushmiPullyu::AIP::SolrFetcher.get_permission_object_ids(@noid)

# which calls this method

def self.get_permission_object_ids(noid)
   # Return array of ids
   new.get_permission_object_ids(noid)
 end

```

cwant: Done.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -0,0 +1,215 @@
...
+    PushmiPullyu::AIP.validate(noid)
+  end
+
+  def run
+    make_directories
+    PushmiPullyu.logger.info("#{@noid}: Retreiving data from Fedora ...")
+
+    [:main_object, :fixity, :content_datastream_metadata, :versions, :thumbnail,
+     :characterization, :fedora3foxml, :fedora3foxml_metadata].each do |item|
+      path_spec = aip_paths.send(item)
```
murny: Instead of using ruby`s metaprogramming `send`, openstruct supports the `[](name)` notation. So this can simply be `aip_paths[item]` 

http://ruby-doc.org/stdlib-2.0.0/libdoc/ostruct/rdoc/OpenStruct.html#5B-5D-method

lib/pushmi_pullyu/aip/creator.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def initialize(noid)
+    @noid = noid
+    @aip_directory = PushmiPullyu::AIP.aip_directory(noid)
+    @aip_filename = PushmiPullyu::AIP.aip_filename(noid)
+    PushmiPullyu::AIP.validate(noid)
+  end
+
+  def run(should_skip_download: false, should_clean_work_directories: true)
+    download_aip unless should_skip_download
```
murny: I would probably decouple the download process away from the creator class. You can call this in the CLI run loop.
Something like this:
```ruby
def run_tick_loop
#....
PushmiPullyu::AIP.download(item) if should_skip_download?
aip_filename = PushmiPullyu::AIP.create(item)
#...
```
Then you can remove the arguements for should_skip_download/and the `download_aip` method in here.

This will give something using the  `PushmiPullyu::AIP.download` method as well since it is currently never being used anywhere else.

lib/pushmi_pullyu/aip/downloader.rb
cwant: Thanks, I didn't know that. This change has been pushed.

lib/pushmi_pullyu/aip/creator.rb
cwant: This is done: the downloaded and the creator are decoupled and the 'should_skip_download' option is removed (I know, I said I wouldn't work on this any more this sprint, but this didn't take very long).

I also added some test fixture data to mock previously downloaded AIP data for the creator tests to work on.

murny: Thanks chris! Looking pretty good. Probably give it a final review and merge it in today

PR #54, opened 2017-05-10
#47: Fixes for Handle PMPY configuration via yaml file

murny:
Make queue top level option and fix option setter to be a merge on default options instead of a straight up assignment/override.


lib/pushmi_pullyu.rb
```diff
@@ -50,6 +50,10 @@ def self.options
   end
 
   def self.options=(opts)
+    options.merge!(opts)
```
mbarnett: 🎉

Issue #26, opened 2017-04-24
Uploading AIP to Swift


murny:
PUT request for uploading AIP to Swift using it's REST API with checksum of bag (using swift api auth token).

No idea how Swift works, so requires some investigation with using it's API for uploading files. Use a ruby http client API to do these requests. Try to use the builtin Ruby libraries preferably. 

Swift REST API documentation: https://developer.openstack.org/api-ref/object-storage/
Chunking of large files (>5gb): https://docs.openstack.org/developer/swift/api/large_objects.html

https://github.com/ruby-openstack/ruby-openstack is probably best Gem out there for this. It also supports etag/chunking of large files as well.

Example script: https://gist.github.com/murny/c4696ec251ec43214303d218d525837b

murny:
Any updates on this @astrilet?  Need help with anything?

murny:
@astrilet can you update this ticket please? 

Maybe create a branch and push your code up so people can see where you are at with this and provide help if you need it? 

You think you will be finished this by early next week? (monday/tuesday)?

murny:
Hey @astrilet, before I can close this and consider this done. I need some documentation on how we would support uploading 5 GB+ files to swift. I remember in standups you mention playing around with this. Can you add some documentation to this ticket on how this would look like if we were to implement this?

I believe `ruby-openstack gem` has support for this via chunking data steams and supplying a manifests. 

Thanks.

astrilet:
swift has a limit of upload file size 5 GB. If you have file larger then 5 GB it needs to be broken into smaller pieces. Python client, for example, does it for you. If we need to upload files > 5GB we need to implement logic to break file into chunks as well. I had some code to do that  but as I understood we decided not to implement it right now and leave it as a future release feature.

murny:
@astrilet would you be able to add this code of yours that you have into this ticket? Or maybe create a gist with it?

This will help everyone for when we want to implement this functionality in the future. Thanks.

astrilet:
OK will create gist and share some code

astrilet:
<script src="https://gist.github.com/astrilet/b383667de73ffb3c7560e53f24411583.js"></script>

astrilet:
Code was committed in PR#57

murny:
Can you make both code snippets into a single gist? and If dynamic is the preferred way lets just use that way (aka remove anything relating to the static manifest code stuff).

[...]

PR #57, opened 2017-05-16
Adding Swift component 

astrilet:


lib/pushmi_pullyu/cli.rb
```diff
@@ -25,7 +25,7 @@ def parse(argv = ARGV)
     PushmiPullyu.options = opts
   end
 
-  def run
+  def run version = ">= 0.a"
```
murny: is this a mistake? Shouldn't need to pass a version param here?

murny:
Hey @astrilet, Make sure you run `rubocop` on your branch. It will catch alot of the style issues I am seeing.

lib/pushmi_pullyu/cli.rb
astrilet: yes, do not need that

lib/pushmi_pullyu/cli.rb
```diff
@@ -139,14 +140,17 @@ def rotate_logs
   def run_tick_loop
     while PushmiPullyu.server_running?
       # Preservation (TODO):
-      # 1. Montior queue
+      # 1. Montior queuek
```
murny: oops, accidental 'K' appended at the end of this comment?

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
+require 'openstack'
+
+class PushmiPullyu::SwiftDepositer
+
+  def initialize(connection: nil, container: nil)
```
murny: I would let ruby do it's job here. Instead of passing default arguments for nil, then checking if these arguments are nil and raising your exception. Just don't use default arguments instead?

Your code: 
```
2.4.0 :001 > PushmiPullyu::SwiftDepositer.new
RuntimeError: conection can not be nil
```
Rewritting this to be ```  def initialize(connection, container)``` now becomes:
```
2.4.0 :001 > PushmiPullyu::SwiftDepositer.new
ArgumentError: wrong number of arguments (given 0, expected 2)
```

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+                                                     authtenant_name: tenant,
+                                                     service_type: service_type)
+    @swift_container = container
+    @logger = PushmiPullyu.logger
+  end
+
+  def deposit_file(file_name)
+    # check if file exists
+    raise "File #{file_name} does not exist" unless File.file?(file_name)
+    file_base_name = File.basename(file_name)
```
murny: File.basename will also return the extension. So you may want to filter this out. Unless we want to name our swift objects like "noid12345.tar"

Can probably just do: `file_base_name = File.basename(file_name, '.*')` 

Examples from docs: https://ruby-doc.org/core-2.3.3/File.html#method-c-basename

astrilet: OK, will fix that

pushmi_pullyu.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'daemons', '~> 1.2', '>= 1.2.4'
   spec.add_runtime_dependency 'redis', '~> 3.3'
   spec.add_runtime_dependency 'connection_pool', '~> 2.2'
+  spec.add_runtime_dependency 'openstack', '~> 3.3.10'
```
murny: bundler will give us a warning here for dependencies that are too tight. Relax this to simply this:

` spec.add_runtime_dependency 'openstack', '~> 3.3'` 

If you need 3.3.10 or above then do 

spec.add_runtime_dependency 'openstack', '~> 3.3', '>= 3.3.10'

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+    tenant       = connection[:tenant]
+    endpoint     = connection[:endpoint]
+    service_type  = 'object-store'
+    auth_method   = 'password'
+    auth_version = connection[:auth_version]
+
+    @swift_connection = OpenStack::Connection.create(username: user,
+                                                     api_key: pass,
+                                                     auth_method: auth_method,
+                                                     auth_url: "#{endpoint}/auth/#{auth_version}",
```
murny: Would it make sense to just have an `auth_url` option in the defaults? Instead of having both `endpoint`/`auth_version` ones?

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+  def deposit_file(file_name)
+    # check if file exists
+    raise "File #{file_name} does not exist" unless File.file?(file_name)
+    file_base_name = File.basename(file_name)
+
+    # calculate file hash
+    hash = Digest::MD5.file(file_name).hexdigest
+
+    # check that container exits
+    raise "Container #{@swift_container} does not exist" unless @swift_connection.container_exists?(@swift_container)
```
murny: Maybe just call `era_container = @swift_connection.container(@swift_container)` or something and check if era_container is an error/nil (exception is raised etc)? Otherwise your doing two API calls here. One for checking if it exists and another for actually going and grabbing the container.

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+    auth_version = connection[:auth_version]
+
+    @swift_connection = OpenStack::Connection.create(username: user,
+                                                     api_key: pass,
+                                                     auth_method: auth_method,
+                                                     auth_url: "#{endpoint}/auth/#{auth_version}",
+                                                     authtenant_name: tenant,
+                                                     service_type: service_type)
+    @swift_container = container
+    @logger = PushmiPullyu.logger
```
mbarnett: I would get rid of this cache variable. Just use `PusmiPullyu.logger` anywhere you're currently using `@logger`. It will simplify reading the code.

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
+require 'openstack'
+
+class PushmiPullyu::SwiftDepositer
+
+  def initialize(connection: nil, container: nil)
+    raise 'conection can not be nil' if connection.nil?
+    raise 'container can not be nil' if container.nil?
+
+    user         = connection[:username]
```
murny: These variable assignments are not needed. Just plug them into the openstack connect on line 17. There never used again.

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
+require 'openstack'
+
+class PushmiPullyu::SwiftDepositer
+
+  def initialize(connection: nil, container: nil)
+    raise 'conection can not be nil' if connection.nil?
```
mbarnett: `conection` is misspelled in the error message. The error message should also be capitalized here and on the next line

mbarnett: But as Shane said, you should get rid of both of these lines, as you're just trying to do Ruby's job for it.

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+    file_base_name = File.basename(file_name)
+
+    # calculate file hash
+    hash = Digest::MD5.file(file_name).hexdigest
+
+    # check that container exits
+    raise "Container #{@swift_container} does not exist" unless @swift_connection.container_exists?(@swift_container)
+
+    # get container object
+    era_container = @swift_connection.container(@swift_container)
```
mbarnett: The `.container` method throws a `OpenStack::Exception::ItemNotFound` exception if the container does not exist ([as seen here](https://github.com/ruby-openstack/ruby-openstack/blob/9a51ecbd08fe168e3886d0452e0856e38219c4a5/lib/openstack/swift/connection.rb#L25)). You can get rid of the redundant check you're doing on line 36.

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    if era_container.object_exists?(file_base_name)
+      # if file already exists, update it with new data
+      @logger.debug("File object #{@swift_container}/#{file_base_name} aleary in the swift, updating content")
```
mbarnett: Typo: `alreary`. Please consider running a spellchecker on your work, there have been a lot of these.

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+      deposited_file = era_container.object(file_base_name)
+      deposited_file.write(File.open(file_name),
+                           etag: hash, content_type: 'application/octet-stream')
+    else
+      # create new deposit file
+      @logger.debug("Creating new file object #{@swift_container}/#{file_base_name} in the swift")
+      deposited_file = era_container.create_object(file_base_name,
+                                                   { etag: hash, content_type: 'application/octet-stream' },
+                                                   File.open(file_name))
```
mbarnett: This isn't very DRY. You're repeating the content_type and File opening logic on line 45 and here. Both `.object` and `.create_object` return an instance of `StorageObject` that responds to `.write`, so you can simplify this to simply:

```ruby
   deposited_file = if era_container.object_exists?(file_base_name)
     era_container.object(file_base_name)
   else
     era_container.create_object(file_base_name)
   end
   deposited_file.write(File.open(file_name),  etag: hash, content_type: 'application/octet-stream')
```

lib/pushmi_pullyu/swift_deposit.rb
```diff
@@ -0,0 +1,58 @@
...
+  def initialize(connection: nil, container: nil)
+    raise 'conection can not be nil' if connection.nil?
+    raise 'container can not be nil' if container.nil?
+
+    user         = connection[:username]
+    pass         = connection[:password]
+    tenant       = connection[:tenant]
+    endpoint     = connection[:endpoint]
+    service_type  = 'object-store'
+    auth_method   = 'password'
```
mbarnett: `service_type` and `auth_method` don't need to be local variables, as they are never changed, and are only read from once. You can move these string into the `OpenStack::Connection.create` call directly, or replace the variables with constant declarations.

lib/pushmi_pullyu/swift_deposit.rb
astrilet: we have different authentication mechanisms (config file and keystone) hence auth_version

lib/pushmi_pullyu/swift_deposit.rb
astrilet: good point will fix that

lib/pushmi_pullyu/swift_deposit.rb
murny: Random side question but isn't there a 'application/tar' MIME type? Should we use that instead?

lib/pushmi_pullyu/swift_deposit.rb
astrilet: will fix

astrilet: will fix

astrilet: will fix

pushmi_pullyu.gemspec
astrilet: OK

lib/pushmi_pullyu/swift_deposit.rb
mbarnett: Other possibilities seem to include `application/x-tar` and `application/x-gtar` in the specific case of GNU Tar. @pbinkley Thoughts? This isn't covered in your original bash mockup.

Octet-stream seems safe but less than idea in its vagueness.

pbinkley: We may as well be as precise as we know how, so if we're using gnu tar (not sure what's on the Centos boxes) let's say so, otherwise say `application/x-tar`. Anyone who's puzzled will have the `.tar` on the file name to help them sort it out.

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  it 'valid connection established' do
+    swift_connection = described_class.new({ username: 'test:tester',
+                                             password: 'testing',
+                                             tenant: 'tester',
+                                             endpoint: 'http://127.0.0.1:8080',
+                                             auth_version: 'v1.0' },
+                                           'ERA')
+    expect(swift_connection).not_to be_nil
```
mbarnett: This isn't a useful test. The only way `swift_connection` will not be nil is if `.new` threw an exception.

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    expect(swift_connection).to be_an_instance_of(described_class)
```
mbarnett: This isn't a useful test. You called `described_class.new`, so ipso facto `swift_connection` _must_ be an instance of `described_class` unless Ruby itself is broken

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    expect(swift_connection.swift_connection).to be_an_instance_of(OpenStack::Swift::Connection)
```
mbarnett: This is just testing that the Gem does what it says on the tin. Assume that the Gem itself is well tested and does what its documentation says it does. It is not useful to rewrite the Gem's tests.

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+RSpec.describe PushmiPullyu::SwiftDepositer do
+  it 'not valid without arguments' do
+    expect { described_class.new }.to raise_error(ArgumentError)
```
mbarnett: This is just testing that Ruby isn't broken. You should assume that Ruby works as Ruby does, and not write tests to validate that the universe isn't broken.

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -0,0 +1,55 @@
...
+    file_base_name = File.basename(file_name)
+
+    # calculate file hash
+    hash = Digest::MD5.file(file_name).hexdigest
+
+    # get container object
+    era_container = @swift_connection.container(@swift_container)
+
+    # check that container exits
+    raise "Container #{@swift_container} does not exist" if era_container.nil?
```
mbarnett: Why would this be `nil`? As noted in a previous review, the `.container` call *raises an exception* if the container doesn't exist. It is impossible for this to be nil given the OpenStack gem's code

mbarnett:
Hey Alex, just a reminder that there are some review comments hidden in the "outdated" buttons above that you still haven't fixed. For your convenience, there are the ones that still need fixing:

https://github.com/ualbertalib/pushmi_pullyu/pull/57#discussion_r116821367

https://github.com/ualbertalib/pushmi_pullyu/pull/57#discussion_r116821164

https://github.com/ualbertalib/pushmi_pullyu/pull/57#discussion_r116826065

lib/pushmi_pullyu/swift_deposit.rb
astrilet: K, will change that

spec/pushmi_pullyu/swift_depositer_spec.rb
astrilet: k will remove that

lib/pushmi_pullyu/swift_deposit.rb
astrilet: will change that

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -0,0 +1,50 @@
...
+
+class PushmiPullyu::SwiftDepositer
+
+  attr_reader :swift_connection, :swift_container
+
+  def initialize(connection, container)
+    raise 'conection can not be nil' if connection.nil?
+    raise 'container can not be nil' if container.nil?
+
+    endpoint     = connection[:endpoint]
```
murny: just plug endpoint/auth_version into the auth_url down on line 17. No need to create these variables

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -0,0 +1,50 @@
...
+  def deposit_file(file_name)
+    # check if file exists
+    raise "File #{file_name} does not exist" unless File.file?(file_name)
+    file_base_name = File.basename(file_name, '.*')
+
+    # calculate file hash
+    hash = Digest::MD5.file(file_name).hexdigest
+
+    # get container object
+    era_container = @swift_connection.container(@swift_container)
```
murny: You created an attr_reader for swift_connection. Use the getter?

murny: Same thing about anywhere `@swift_container` is used. Use the getter `swift_container`

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -0,0 +1,50 @@
+require 'openstack'
+
+class PushmiPullyu::SwiftDepositer
+
+  attr_reader :swift_connection, :swift_container
+
+  def initialize(connection, container)
+    raise 'conection can not be nil' if connection.nil?
```
mbarnett: Type: 'connection'

astrilet: sorry, will fix

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  it 'deposit file' do
+    VCR.use_cassette('swift_deposit') do
+      swift_connection = described_class.new({ username: 'test:tester',
+                                               password: 'testing',
+                                               tenant: 'tester',
+                                               endpoint: 'http://www.example.com:8080',
+                                               auth_version: 'v1.0' },
+                                             'ERA')
+
+      sample_file = "#{File.dirname(__FILE__)}/../fixtures/config.yml"
```
murny: you can just do `sample_file = 'spec/fixtures/config.yml' `. Much nicer then having in my opinion.

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'spec_helper'
+
+RSpec.describe PushmiPullyu::SwiftDepositer do
+  it 'not valid with nil arguments' do
+    expect { described_class.new(nil, nil) }.to raise_error(RuntimeError)
+  end
+
+  it 'valid connection established' do
```
murny: You could pretty much combine this spec with the one below. As they are very similar. Both are creating an instance of PushmiPullyu::SwiftDepositer. Can just move the `expect()` from this one down into the other spec. Then everything you wanted to test is still done, without all this extra code (and another VCR tape).

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+                                             'ERA')
+      expect(swift_connection).not_to be_nil
+      expect(swift_connection).to be_an_instance_of(described_class)
+      expect(swift_connection.swift_connection).to be_an_instance_of(OpenStack::Swift::Connection)
+    end
+  end
+
+  it 'deposit file' do
+    VCR.use_cassette('swift_deposit') do
+      swift_connection = described_class.new({ username: 'test:tester',
```
murny: `swift_connection` is a misleading variable name. This should be `swift_depositer` instead. (as `swift_depositer.swift_connection` is the actual swift_connection)

lib/pushmi_pullyu/swift_depositer.rb
astrilet: OK, will do

lib/pushmi_pullyu/swift_depositer.rb
astrilet: will fix

spec/pushmi_pullyu/swift_depositer_spec.rb
astrilet: will do

spec/pushmi_pullyu/swift_depositer_spec.rb
astrilet: will rename

lib/pushmi_pullyu/cli.rb
```diff
@@ -160,15 +160,21 @@ def run_tick_loop
     while PushmiPullyu.server_running?
       # Preservation (TODO):
       item = @queue.wait_next_item
-      logger.debug(item)
-      # 3. Retrieve GenericFile data in fedora
-      # 4. creation of AIP
-      # 5. bagging and tarring of AIP
-      # 6. Push bag to swift API
-      file_to_deposit = './pushmi_pullyu'
```
murny: Bad Merge? You removed all your code?

astrilet: yes, I removed that  I thought we will need to change that anyways... can add it back

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -1,11 +1,9 @@
 require 'spec_helper'
 
 RSpec.describe PushmiPullyu::SwiftDepositer do
-  it 'not valid with nil arguments' do
+  it 'valid connection established' do
     expect { described_class.new(nil, nil) }.to raise_error(RuntimeError)
-  end
 
-  it 'valid connection established' do
```
murny: Theses were the wrong two tests to merge haha. The first one (not valid with nil arguments) is fine. It's the other two as your testing the same thing twice.

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -1,11 +1,9 @@
 require 'spec_helper'
 
 RSpec.describe PushmiPullyu::SwiftDepositer do
-  it 'not valid with nil arguments' do
```
murny: Oops you merged the wrong two tests. This one ( 'not valid with nil arguments') was completely fine. It was the other two tests in this file. They can be merged as they are testing the exact same pieces of code. Hence you can merge them together and have less code as a result (including one less VCR file)

murny:
@astrilet see this comment on the ticket with regards to doing some research and adding some documentation on how to support big upload sizes to swift (5 GB+ size files): https://github.com/ualbertalib/pushmi_pullyu/issues/26#issuecomment-302542883

After this is done, ill close out this ticket.

PR #59, opened 2017-05-16
Add basic error reporting to PMPY

### LINKED ISSUE #32: Add application monitoring for PushmiPullyu
Bring in something like ~~newrelic/skylight~~ (can't use for pushmipullyu since there is no webstack?) errbit  (self hosted error tracker based on airbrake), or honeybadger (which apparently is free for non-commerical open source projects?)/etc for PushmiPullyu so we can get some eyes on how PushmiPullyu is performing when in production and being alerted on failures/issues.

mbarnett:
Basic reporting of errors to Rollbar. The example run-loop setup will report the error and crash the daemon if an error occurs while retrieving a noid from the queue, or if the noid is retrieved successfully, report any errors that occur in the loop with information about the current noid, but continue on to the next noid. We can tweak this as we like.

Rollbar integration is only active when a token is passed in the options. Rollbar is configured to not pass errors back to the API even with a token configured, if RSpec is running.

I've had to do a localized cop disable to make rubocop happy, but that may indicate that we don't want that cop.

Closes #32

### DIFF
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -1,6 +1,7 @@
 require 'erb'
 require 'fileutils'
 require 'optparse'
+require 'rollbar'
 require 'singleton'
 require 'yaml'
 
@@ -26,12 +27,19 @@ class PushmiPullyu::CLI
   end
 
   def run
-    if options[:daemonize]
-      start_server_as_daemon
-    else
-      # If we're running in the foreground sync the output.
-      $stdout.sync = $stderr.sync = true
-      start_server
+    configure_rollbar
+    begin
+      if options[:daemonize]
+        start_server_as_daemon
+      else
+        # If we're running in the foreground sync the output.
+        $stdout.sync = $stderr.sync = true
+        start_server
+      end
+    # rubocop:disable Lint/RescueException
+    rescue Exception => e
+      Rollbar.error(e)
+      raise e
     end
   end
 
@@ -47,6 +55,13 @@ class PushmiPullyu::CLI
   end
 
   private
+
+  def configure_rollbar
+    Rollbar.configure do |config|
+      config.enabled = false unless options[:rollbar_token].present?
+      config.access_token = options[:rollbar_token]
+    end
+  end
 
   def options
     PushmiPullyu.options
@@ -77,6 +92,10 @@ class PushmiPullyu::CLI
 
       o.on('-d', '--debug', 'Enable debug logging') do
         opts[:debug] = true
+      end
+
+      o.on('-r', '--rollbar-token TOKEN', 'Enable error reporting to Rollbar') do |token|
+        opts[:rollbar_token] = token if token.present?
       end
 
       o.on '-C', '--config PATH', 'path to YAML config file' do |config_file|
@@ -139,15 +158,22 @@ class PushmiPullyu::CLI
   def run_tick_loop
     while PushmiPullyu.server_running?
       # Preservation (TODO):
-      # 1. Montior queue
-      # 2. Pop off GenericFile element off queue that are ready to begin process preservation event
       item = @queue.wait_next_item
-      logger.debug(item)
-      # 3. Retrieve GenericFile data in fedora
-      # 4. creation of AIP
-      # 5. bagging and tarring of AIP
-      # 6. Push bag to swift API
-      # 7. Log successful preservation event to log files
+
+      # add additional information about the error context to errors that occur while processing this item.
+      Rollbar.scoped(noid: item) do
+        begin
+        # 3. Retrieve GenericFile data in fedora
+        # 4. creation of AIP
+        # 5. bagging and tarring of AIP
+        # 6. Push bag to swift API
+        # 7. Log successful preservation event to log files
+        rescue => e
+          Rollbar.error(e)
+          # TODO: we could re-raise here and let the daemon die on any preservation error, or just log the issue and
+          # move on to the next item.
+        end
+      end
 
       rotate_logs if PushmiPullyu.reset_logger?
     end
--- a/pushmi_pullyu.gemspec
+++ b/pushmi_pullyu.gemspec
@@ -24,9 +24,13 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.3.1'
 
   spec.add_runtime_dependency 'activesupport', '~> 5.0'
+  spec.add_runtime_dependency 'connection_pool', '~> 2.2'
   spec.add_runtime_dependency 'daemons', '~> 1.2', '>= 1.2.4'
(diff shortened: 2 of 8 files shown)


pushmi_pullyu.gemspec
```diff
@@ -24,9 +24,10 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.3.1'
 
   spec.add_runtime_dependency 'activesupport', '~> 5.0'
+  spec.add_runtime_dependency 'connection_pool', '~> 2.2'
   spec.add_runtime_dependency 'daemons', '~> 1.2', '>= 1.2.4'
   spec.add_runtime_dependency 'redis', '~> 3.3'
-  spec.add_runtime_dependency 'connection_pool', '~> 2.2'
+  spec.add_runtime_dependency 'rollbar'
```
murny: pin this to a reasonable minor version

mbarnett: I might just pin this to the major. Generally speaking you want to be at the tip of any API fixes for this kind of thing

lib/pushmi_pullyu/cli.rb
```diff
@@ -26,12 +27,19 @@ def parse(argv = ARGV)
...
+    configure_rollbar
+    begin
+      if options[:daemonize]
+        start_server_as_daemon
+      else
+        # If we're running in the foreground sync the output.
+        $stdout.sync = $stderr.sync = true
+        start_server
+      end
+    # rubocop:disable Lint/RescueException
```
murny: Do you agree with any of their alternatives? https://github.com/bbatsov/ruby-style-guide#no-blind-rescues

```
# good
begin
  # a blind rescue rescues from StandardError, not Exception as many
  # programmers assume.
rescue => e
  # exception handling
end

# also good
begin
  # an exception occurs here
rescue StandardError => e
  # exception handling
end
```

mbarnett: The problem is, that will mean certain kinds of non-standarderror exceptions won't get reported to Rollbar. It's good advice in general, just not when you actually do want a failsafe to catch _everything_, so you can report it.

everyone

murny: Have an example off your head that wouldn't be captured by StandardError? I'm in agreeing with leaving it as is. Even though I hate inline rubocop comments.

mbarnett: You can see them all here, https://ruby-doc.org/core-2.3.0/Exception.html but 2 big classes of errors are the SignalEror (for obvs reasons important with our signal handlers), and ScriptError which syntax errors fall under. Always good to know about Production run time typos

murny: Yeah rubocop is a bit silly here. Basically it should only say that if you didnt reraise the exception.

``` # Never rescue Exception *unless* you re-raise in rescue body```

File: spec/pushmi_pullyu/cli_spec.rb [hunk 3]
```diff
@@ -27,6 +27,16 @@
         expect(cli).not_to have_received(:start_server)
       end
     end
+
+    context 'should set up Rollbar' do
+      it 'starts working loop as daemon' do
```
murny: copy pasted description?

spec/spec_helper.rb
```diff
@@ -8,6 +8,13 @@
 # Require support scripts
 Dir[File.dirname(__FILE__) + '/support/**/*.rb'].each { |f| require f }
 
+# disable Rollbar reporting even if a Rollbar token is passed
```
murny: move this to a `spec/support` rb file? To keep this file clean?

mbarnett:
Added the config example. I think VCRing their API is probably a bridge too far, though -- it's their job to test that their gem works with their API, and its the nature of these gems to silently fail if there's a service error, rather than take down the client app. This can probably fall under "we assume the universe is sane".

pushmi_pullyu.gemspec
```diff
@@ -24,9 +24,13 @@ Gem::Specification.new do |spec|
...
 
   spec.add_runtime_dependency 'activesupport', '~> 5.0'
+  spec.add_runtime_dependency 'connection_pool', '~> 2.2'
   spec.add_runtime_dependency 'daemons', '~> 1.2', '>= 1.2.4'
   spec.add_runtime_dependency 'redis', '~> 3.3'
-  spec.add_runtime_dependency 'connection_pool', '~> 2.2'
+  spec.add_runtime_dependency 'rollbar', '~> 2.14'
+
+  # Rollbar recommended json serializer & version
+  spec.add_runtime_dependency 'oj', '~> 2.12.14'
```
murny: `spec.add_runtime_dependency 'oj', '~> 2.12', '>= 2.12.14'`

spec/pushmi_pullyu/cli_spec.rb
mbarnett: N...no... >.>

pushmi_pullyu.gemspec
mbarnett: Per their docs, they want it pinned to 2.12.14.x https://rollbar.com/docs/notifier/rollbar-gem/#getting-started

murny: Old docs? OJ gem is on version 3.0.8 now: 
https://github.com/ohler55/oj
https://rubygems.org/gems/oj

And looks like it follows Sem Ver too

mbarnett: It's also in their Readme, updated less than 2 months ago https://github.com/rollbar/rollbar-gem/blob/master/README.md, and their Gemfile https://github.com/rollbar/rollbar-gem/blob/master/Gemfile, so I don't think they test against anything beyond that. Safer to stay in the tested zone

murny: Yeah not sure. Must be doing something weird here on rollbar's end. Anyways this is fine, leave as is. LGTM.

murny: They had this line in their code:

``` if RUBY_VERSION.start_with?('2.4') && is_not_jruby
30	  gem 'oj', '~> 2.16.0'
```

Also its why travis is blowing up. That version of OJ doesnt work on 2.4 ruby.

mbarnett: They're pinned to that because of a bug in later versions that breaks AR serialization, 2.13 introduced, then fixed, then 2.14 introduced some others, etc, etc.

Unfortunately it looks like we need to drop Ruby 2.4 or go with slower Json generation/error reporting

murny: I honestly vote to go with the slower JSON? Especially when were capped at 5000 requests per month. We don't need that crazy amount of speed. Plus default ruby json serailizer has come along way from what it was.

mbarnett: Yeah, I'm thinking the same

PR #61, opened 2017-05-18
Add danger to project

murny:
Adds https://github.com/danger/danger to projects CI. 

To make PRs even more annoying :trollface:


Dangerfile
```diff
@@ -0,0 +1,25 @@
...
+if has_app_changes && !has_spec_changes
+  warn('There are code changes, but no corresponding tests. '\
+         'Please include tests if this PR introduces any modifications in '\
+         'behavior.',
+       sticky: false)
+end
+
+# Make it more obvious that a PR is a work in progress and shouldn't be merged yet
+warn('PR is classed as Work in Progress') if github.pr_title.include? '[WIP]'
+
```
cwant: Might want to expand this a bit. I have seen Matt use "[DON'T MERGE]" (which implies WIP), and I have used "[WIP -- Don't merge]" which doesn't match the spec.

Something like:

```ruby
[/WIP/, /don\'t merge/i, /work in progress/i ].each do |wip|
  (warn('PR is classed as Work in Progress'); break) if github.pr_title =~ wip
end
```

Dangerfile
```diff
@@ -0,0 +1,25 @@
...
+# just leaving a title
+warn('Please add a detailed summary in the description.') if github.pr_body.length < 5
+
+# Ensure a clean commits history
+if git.commits.any? { |c| c.message =~ /^Merge branch '#{github.branch_for_base}'/ }
+  warn 'Please rebase to get rid of the merge commits in this PR'
+end
+
+# Warn when there is a big PR
+warn('This PR is too big! Consider breaking it down into smaller PRs.') if git.lines_of_code > 1000
```
cwant: Hey c'mon! Nobody will every submit a PR that big!! :roll_eyes: :smile:

Dangerfile
murny: Might just toss this one? It's kinda pointless. Most of these I stole from other Dangerfiles on other projects

Dangerfile
murny: Haha :sweat_smile: .  We can change this to 2000 or something too later. But the basic idea is there (bigger PRs take longer to review, etc). I mean these are all warnings (we can use this stuff to fail builds too if we want to be more strict about it) so just take them as a little friendly nudge.

cwant: I fully agree with what is there, and I wouldn't even mind if it became a stricter fail rule. My apologies for the mondo PR that has been churning for the last couple of weeks, it really should have been broken down into manageable chunks (and it would have if the system had automatically rejected it).

mbarnett: Does 1000 seem large to anyone else? Hydra's (Esme's) guidelines doc says 400 is roughly their threshold. I had 300 in my head as an "ideal size", but I don't know where that idea comes from. I think Chris' recent PR was 700ish (I think?), so if we as a group felt that was too large, 1000 seems like it might be too high.

murny: this was originally 500. And is the standard on a lot of other dangerfiles I have seen. Can change back.

PR #69, opened 2017-05-26
Code cleanup

murny:
Bunch of cleanups of the codebase so far:

- Fix requires everywhere (requiring things we are not using or things we should be requiring that we are using)
- Move custom AIP exceptions to where they are being used in the code
- Remove constants as they are never used anywhere else in codebase
- Add RDF gem as dependency as we are using this (indirectly via rdf-n3 gem)
- Use RDF query instead of looping through every graph node for the specific predicate we are searching for
- Only validate noid on the entry methods of the AIP module
- Remove constants for options hash, never being used anywhere else
- Update rubocop to latest (white space styles are now under layout namespace)
 - Use `minitar` instead of `archive/tar/minitar` as latter is deprecated
- Have classes do their one job (AIP download downloads, AIP creator creates, AIP destroy cleanups, previously creator was downloading and cleaning up which is redundant)
- General Code cleanup and refactors:
- No need to error check things like swift connection is nil (it can never be nil you are always passing in a hash) or the container is nil (if it is nil, the code will raise an exception when it tries to use it, let ruby do it's job)
  - use getters
  - remove code that is never used
  - remove variables that are never used more then once
  - use predicate verbs for booleans and have things return booleans instead of numbers/strings
  - privatize methods that shouldn't be public
  - use instance variables from class intialization instead of passing them again as arguments
  - simplfy dynamic code arguments, fix one offs cases
  - organize methods/dependencies/etc alphabetically from A-Z where it makes sense
  - etc
- General Cleanup for specs:
  - Make sure we are testing branching logic
  - Add missing unit tests for classes/methods
  - Don't wrap everything in VCR blocks, only things that are needed and making API calls
  - Unit tests should be testing the units of the code, not be integration tests
  - copy pasted of spec setup  only use what you need in your specs
  - test what your code is actually doing


lib/pushmi_pullyu/aip.rb
```diff
@@ -1,63 +1,36 @@
-require 'pushmi_pullyu'
-require 'bagit'
-require 'archive/tar/minitar'
+require 'fileutils'
 
 module PushmiPullyu::AIP
-  # Exceptions
-  class BagInvalid < StandardError; end
-  class NoContentFilename < StandardError; end
-  class FedoraFetchError < StandardError; end
```
murny: Move these to where they are being used. Helps with maintainability. If we decide to remove FedoraFetcher class one day, this doesn't get orphaned and forgotten about.

lib/pushmi_pullyu/aip/creator.rb
```diff
@@ -1,65 +1,40 @@
-require 'archive/tar/minitar'
```
murny: `archive/tar/minitar` is apparently depreacted. Use `minitar` instead.

https://github.com/halostatue/minitar#description

lib/pushmi_pullyu/aip/creator.rb
```diff
@@ -1,65 +1,40 @@
...
 
   def initialize(noid)
     @noid = noid
     @aip_directory = PushmiPullyu::AIP.aip_directory(noid)
     @aip_filename = PushmiPullyu::AIP.aip_filename(noid)
-    PushmiPullyu::AIP.validate(noid)
   end
 
-  def run(should_clean_work_directories: true)
+  def run
```
murny: Heavily simplify this. Creator class, should do just that, create the aip file (bagging and tarring). We don't need to do the cleanup of the directories or filename. We have a `PushmiPullyu::AIP.destroy(noid)` that is called that does exactly that. Also you don't need to check if the tar already exists and remove it. The tarring process will override an existing tar file if it already exists.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
+require 'fileutils'
+require 'ostruct'
+require 'rdf'
```
murny: We are using RDF in this class (not via rdf/n3) lets require this and bring in the gem as a dependency (incase rdf/n3 does something weird one day)

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
-
 class PushmiPullyu::AIP::Downloader
 
-  FILENAME_PREDICATE = 'info:fedora/fedora-system:def/model#downloadFilename'.freeze
+  class NoContentFilename < StandardError; end
 
   def initialize(noid)
     @noid = noid
-    @fedora_fetcher = PushmiPullyu::AIP::FedoraFetcher.new(@noid)
-    PushmiPullyu::AIP.validate(noid)
```
murny: the validations are all done before we get here. So we are validating this far more then we need to.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
   private
 
-  def download_and_log(path_spec, local_path: nil, fedora_fetcher: nil)
-    # Fetcher is either for current object, or for a permission object
-    fedora_fetcher ||= @fedora_fetcher
-
-    # Sometimes we don't know filename in advance, so we use local_path ...
-    output_file = full_local_path(local_path || path_spec.local)
-    url = fedora_fetcher.object_url(path_spec.remote)
+  def download_and_log(path_spec, fedora_fetcher)
```
murny: Simplify this. There is only one time that the local_path is nil, which is when we need to get the content_filename. This can be done smarter and fix by including it in the path_spec which we are passing in. Also there is only one time we are passing in the fedora_fetcher which is for permissions. Instead of doing a nil check and falling back to the default one, let's just pass it in every time. Much simpler API and no more nil checks everywhere.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
+                                                is_rdf: is_rdf)
+    log_saved(is_success, output_file)
   end
 
   def download_permissions
     PushmiPullyu.logger.info("#{@noid}: looking up permissions from Solr ...")
     results = PushmiPullyu::AIP::SolrFetcher.new(@noid).fetch_permission_object_ids
     if results.empty?
       PushmiPullyu.logger.info("#{@noid}: permissions not found")
-      return
```
murny: just if else this and let it fall through and return naturally.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
-    output_file = full_local_path(local_path || path_spec.local)
-    url = fedora_fetcher.object_url(path_spec.remote)
+  def download_and_log(path_spec, fedora_fetcher)
+    output_file = path_spec.local
 
-    log_fetching(path_spec, url, local_path: output_file)
+    log_fetching(fedora_fetcher.object_url(path_spec.remote), output_file)
 
-    rdf = (output_file =~ /\.n3$/)
+    is_rdf = (output_file !~ /\.n3$/)
```
murny: if we want a boolean, lets use the regex `!~` which returns a boolean instead of a number. Predicate the variable names too that are booleans.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
   def download_permission(permission_id)
-    permission_fetcher =
-      PushmiPullyu::AIP::FedoraFetcher.new(permission_id)
-
-    filename = permission_filename(permission_id)
-
     path_spec = OpenStruct.new(
       remote: nil,
-      local: filename,
+      local: "#{aip_dirs.metadata}/permission_#{permission_id}.n3",
```
murny: Make this like the rest of them. No need for a seperate method.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
+    download_and_log(path_spec, PushmiPullyu::AIP::FedoraFetcher.new(permission_id))
   end
 
   ### Logging
 
-  def log_fetching(path_spec, url, local_path: nil)
-    local_path ||= path_spec.local
-    message = "#{@noid}: #{local_path} -- fetching from #{url} ..."
-    PushmiPullyu::Logging.log_aip_activity(@noid, message)
+  def log_fetching(url, output_file)
```
murny: Simplfy both of these logging methods. No need for `local_path` anymore as `path_spec.local` will always have what we want. Reduce a lot of logic as a result.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
 
   ### Directories
 
   def aip_dirs
     @aip_dirs ||= OpenStruct.new(
-      objects: 'objects',
-      metadata: 'objects/metadata',
-      logs: 'logs',
-      thumbnails: 'thumbnails'
+      objects: "#{aip_directory}/data/objects",
```
murny: we explode these via a `full_local_path` method in many places throughout this code. Just do it here once and don't have to worry about it again.

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
-  end
-
   ### Files
 
   def aip_paths
     @aip_paths ||= OpenStruct.new(
       main_object: OpenStruct.new(
         remote: nil, # Base path
-        local: "#{@aip_dirs.metadata}/object_metadata.n3",
+        local: "#{aip_dirs.metadata}/object_metadata.n3",
```
murny: use the getter instead

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
-    graph.each_statement do |statement|
-      if statement.predicate == FILENAME_PREDICATE
-        @content_filename = "#{aip_dirs.objects}/#{statement.object}"
-        break
-      end
+    filename_predicate = RDF::URI('info:fedora/fedora-system:def/model#downloadFilename')
+
+    graph = RDF::Graph.load(aip_paths.main_object.local)
+
+    graph.query(predicate: filename_predicate) do |results|
```
murny: use RDF query instead of looping through every node of the graph. More efficient and cleaner.

lib/pushmi_pullyu/aip/fedora_fetcher.rb
```diff
@@ -1,65 +1,59 @@
...
 
-    request['Accept'] = RDF_FORMAT if rdf
+    request['Accept'] = RDF_FORMAT if is_rdf
 
     response = Net::HTTP.start(uri.hostname, uri.port) do |http|
       http.request(request)
     end
 
     if response.is_a?(Net::HTTPSuccess)
-      if download_path
```
murny: This is weird. If we have no download_path, this if statement should be at the very start of the method. Why go hit the API if we arent going to do anything with the results? Let's just make this argument manatory instead of optional. Looks like this was done for testing. But as a result, your not even testing what this method is even doing then in a real world setting.

lib/pushmi_pullyu/aip/fedora_fetcher.rb
```diff
@@ -1,65 +1,59 @@
...
       return false
+    else
+      raise FedoraFetchError
     end
+  end
+
+  private
 
-    raise PushmiPullyu::AIP::FedoraFetchError
+  def pairtree
```
murny: Make these methods private. Shouldn't need to be exposed.

lib/pushmi_pullyu/aip/solr_fetcher.rb
```diff
@@ -1,47 +1,33 @@
...
+  class SolrFetchError < StandardError; end
+
   def initialize(noid)
     @noid = noid
   end
 
   def fetch_permission_object_ids
-    json = run_query_json("accessTo_ssim:#{@noid}", fields: 'id', url_extra: nil)
-    hash = JSON.parse(json)
+    hash = JSON.parse(run_query_json)
```
murny: Heavily simpfy this. Don't need to string interpolate most of this. And most of this stuff will never change like fileds `id`, as this is the only place this method is being called. Use instance variable noid instead of passing it in as an arugment.

lib/pushmi_pullyu/aip/solr_fetcher.rb
```diff
@@ -1,47 +1,33 @@
...
-    url += url_extra if url_extra
-    uri = URI(url)
-
-    request = Net::HTTP::Get.new(uri)
-
-    response = Net::HTTP.start(uri.hostname, uri.port) do |http|
-      http.request(request)
-    end
+  # Return fetched results, else raise an error
+  def run_query_json
```
murny: Reduced majority of this code down to 3 lines. Much easier to read and understand now as a bonus.

Using Net:HTTP build in `get_response` method as well since were just doing a simple get request for the response and nothing else.

lib/pushmi_pullyu/logging.rb
```diff
@@ -23,8 +23,8 @@ def self.logger
     @logger ||= initialize_logger
   end
 
-  def self.log_aip_activity(noid, message)
-    log_file = "#{PushmiPullyu.options[:workdir]}/#{noid}/data/logs/aipcreation.log"
+  def self.log_aip_activity(aip_directory, message)
+    log_file = "#{aip_directory}/data/logs/aipcreation.log"
```
murny: just pass in the aip_directory (which is well known and already calculated when calling this) instead of computing the aip_directory once again here.

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -1,13 +1,11 @@
+require 'digest/md5'
 require 'openstack'
 
 class PushmiPullyu::SwiftDepositer
 
-  attr_reader :swift_connection, :swift_container
-
-  def initialize(connection, container)
-    raise 'connection can not be nil' if connection.nil?
```
murny: connection can never be nil. We are always passing in a hash to this method. Even if the hash elements consist of nils like so:
```
{ username: nil, api_key: nil, etc: nil }
```
This is technically still not a nil object.

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -1,13 +1,11 @@
...
-    raise 'container can not be nil' if container.nil?
```
murny: If container is nil, this will raise an expection when we attempt to use it down on line 24. Let ruby do its job and raise the proper exception instead of our custom one.

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -16,30 +14,18 @@ def initialize(connection, container)
...
       service_type: 'object-store'
     )
-
-    @swift_container = container
   end
 
-  def deposit_file(file_name)
-    # check if file exists
-    raise "File #{file_name} does not exist" unless File.file?(file_name)
+  def deposit_file(file_name, swift_container)
```
murny: Moved the container argument into the deposit_file method. This gives a bit more flexibility in case you wanted to deposit files into different containers. You can then use the same SwiftDepositer connection instance instead of creating multiple SwiftDepositer connections for this.

spec/pushmi_pullyu/aip/creator_spec.rb
```diff
@@ -2,25 +2,15 @@
...
-                base_path: '/dev',
-                user: 'fedoraAdmin',
-                password: 'fedoraAdmin' },
-      solr: { url: 'http://www.example.com:8983/solr/development' } }
-  end
   let(:noid) { '9p2909328' }
-  let(:aip_file) { "#{workdir}/#{noid}.tar" }
-  let(:mock_download_data) { "spec/fixtures/aip_download/#{noid}" }
+  let(:aip_file) { "#{aip_folder}.tar" }
+  let(:aip_folder) { "#{workdir}/#{noid}" }
```
murny: Add a aip_folder variable. Here we can use these variables instead of expecting the hardcoded "tmp/creator_spec/9p2909328.tar" or "tmp/creator_spec/9p2909328" multiple times down below. If we ever decide to change theses, its now a one line code change instead of multiple lines.

spec/pushmi_pullyu/aip/creator_spec.rb
```diff
@@ -2,25 +2,15 @@
 
 RSpec.describe PushmiPullyu::AIP::Creator do
   let(:workdir) { 'tmp/creator_spec' }
-  let(:options) do
-    { workdir: workdir,
-      fedora: { url: 'http://www.example.com:8983/fedora/rest',
```
murny: Only create variables for what your using in your tests. Creator class shouldn't need fedora/solr information.

spec/pushmi_pullyu/aip/creator_spec.rb
```diff
@@ -2,25 +2,15 @@
...
-  end
   let(:noid) { '9p2909328' }
-  let(:aip_file) { "#{workdir}/#{noid}.tar" }
-  let(:mock_download_data) { "spec/fixtures/aip_download/#{noid}" }
+  let(:aip_file) { "#{aip_folder}.tar" }
+  let(:aip_folder) { "#{workdir}/#{noid}" }
   let(:creator) { PushmiPullyu::AIP::Creator.new(noid) }
 
   before do
-    allow(PushmiPullyu.logger).to receive(:info)
```
murny: No logging happens in this class, no need to mock them.

spec/pushmi_pullyu/aip/creator_spec.rb
```diff
@@ -29,91 +19,68 @@
   end
 
   describe '#run' do
-    it 'creates the aip, removes work directory by default' do
-      VCR.use_cassette('aip_downloader_run') do
```
murny: Creator class doesn't make API calls. No need to wrap all these in VCR blocks

spec/pushmi_pullyu/aip/solr_fetcher_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
+
+RSpec.describe PushmiPullyu::AIP::SolrFetcher do
```
murny: Missing solr_fetcher unit tests, added this in. (of course its being tests via downloader class, but doesn't hurt to unit test this itself)

spec/pushmi_pullyu/aip_spec.rb
```diff
@@ -1,69 +1,75 @@
...
-    FileUtils.mkdir_p(workdir)
-    FileUtils.cp_r(mock_download_data, workdir)
+    allow(PushmiPullyu).to receive(:options) { { workdir: workdir } }
   end
 
-  after do
-    FileUtils.rm_rf(workdir)
-    FileUtils.rm_rf(aip_file)
-  end
+  describe '.create' do
```
murny: Changed most of these. The actual work of these are being done in the creator/downloader classes, which we are already testing. Here we just want to see if, we are calling instancing those classes and running them, and that we are returning the filename/directory as expected.

spec/pushmi_pullyu/logging_spec.rb
```diff
@@ -28,8 +28,8 @@ class LoggerTest
     expect(PushmiPullyu::Logging.logger).to have_received(:debug).with(an_instance_of(String)).once
   end
 
-  describe '#reopen' do
-    let!(:tmp_dir) { 'tmp/test_dir' }
+  describe '.reopen' do
```
murny: `#` is for instance level methods. Most of these are class/module level methods so should be using `.`

http://www.betterspecs.org/#describe

lib/pushmi_pullyu/aip.rb
```diff
@@ -1,63 +1,36 @@
...
   class NoidInvalid < StandardError; end
 
-  def self.create(noid, should_clean_work_directories: true)
-    # Note: returns the filename of the created AIP tarball
-    validate(noid)
-    PushmiPullyu::AIP::Creator
-      .new(noid)
-      .run(should_clean_work_directories: should_clean_work_directories)
+  def self.create(noid)
+    validate_noid(noid)
```
mbarnett: Do we need to be doing it on each of these?

lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -1,215 +1,184 @@
...
-    local_path ||= path_spec.local
-    message = if success
-                "#{@noid}: #{local_path} -- saved"
+  def log_saved(is_success, output_file)
+    message = if is_success
+                "#{@noid}: #{output_file} -- saved"
               else
-                "#{@noid}: #{local_path} -- not_found"
+                "#{@noid}: #{output_file} -- not_found"
               end
```
mbarnett:
```
message = "#{@noid}: #{output_file} -- " + is_success ? "saved" : "not_found"
```

lib/pushmi_pullyu/aip.rb
```diff
@@ -1,63 +1,36 @@
...
+    PushmiPullyu::AIP::Creator.new(noid).run
+    aip_filename(noid)
   end
 
   def self.download(noid)
```
mbarnett: Personally, I'd ditch the redundant `self.` on all these methods and just use a `module_functions` declaration at the end of each file. Not a big deal, though

lib/pushmi_pullyu/aip.rb
murny: Yeah forsure, much agree. Don't like this currently either as you be validating the noid multiple times. Will refactor to your block idea specified in main review comment. Then we don't need to do this.

lib/pushmi_pullyu/aip.rb
murny: Good call, will change.

lib/pushmi_pullyu/aip.rb
murny: Made these refactors. Much improved. Reduced this class from 50+ lines to 20+ lines. And now only has one access point. Other classes no longer have dependencies on this module either, as I am now passing aip_directory/aip_filename in as arguments.

pbinkley:
Sorry for not getting back to this sooner. Regarding NOIDs: the example @cwant noted in https://confluence.ucop.edu/display/Curation/NOID, "13030/tf5p30086k", is a NOID being used as part of an ARK. The NOID is tf5p30086k - no slashes, just alphanumeric, as @murny noted.  It's not clear to me how strings with slashes came to be treated as NOIDs in this thread. If they're part of the picture, it must be because we're grabbing the wrong thing, not because we're generating NOIDs with slashes in them.

mbarnett:
I think part of the concern was towards engineering towards a solution that could handle valid NOIDs in general, not NOIDs as we happen to generate them at the current time -- the notion being that if we ever tweak the NOID generation pattern in the generator's config files, it is unlikely that we will recall that we would also have to evaluate changing logic in this codebase if this codebase relied on assumptions that were overly familiar with the format in use currently.

The unclarity of the NOID spec and codebase doesn't really help things, here, because as written the active fedora noid generator allows for a NOID with an arbitrary prefix that can include anything, including a slash or any other filesystem- or URL-breaking character. This is, unfortunately, the kind of bad corner-case management that happens when a niche community tries to reinvent a wheel that the broader community solved with UUIDs.

cwant:
The concern in my mind is that we are creating things in the filesystem based on an ill-defined spec that may have slashes either now or in the future.

My personal opinion is that the one or two lines of code needed to sanitize a noid value so that it can live in a filesystem as a single entity (either a directory or as a file by appending '.tar' on the end) is a small investment that buys some peace of mind.

pbinkley:
Fair enough, this *is* the only place I can think of in our ecosystem where we're counting on the NOID being suitable for use as a filename. We'll have the chance to revisit this use in Pushmi-Pullyu when we move to full AIPs and reingest everything, so we've got some slack. There's no harm in future-proofing the code with sanitizing as @cwant suggests. But I can't foresee our changing our NOIDs generation pattern in a way that would cause the problem @mbarnett warns of: the whole point of them is to keep them opaque, and the only reason to fool around with prefixes would be to add meaning.

cwant:
An alternative safeguard: throw NoidInvalid if the noid has a slash in it and let whomever is monitoring Rollbar deal with it. Hopefully this will never happen, but if it does we're covered.

pbinkley:
I think I'm guilty of making two issues intertwingle here: the original
concern about the directory/tarfile name for the AIP in Pushmi-Pullyu, and
the broader issue of primary identifiers that show in the public URL in
HydraNorth, both of which are currently the NOID.

There's no reason to require the AIP directory to be named with the NOID as
I originally spec'ed: once it's buried in the Swift object, the filename is
no more useful than a metadata field for identifying that Swift object. We
could name the directory (and the eventual tar) "aip", and just add the
NOID as a metadata name-value pair (as I think we're already doing) when we
push it into Swift. Swift will provide its own identifier for the object,
and we harvest that identifier back into the Fedora object, so we're
already covered for two disaster scenarios: when we need to recover an
object from PP based on metadata in Fedora, and when we need to
reconstitute the collection from Swift when everything else has burned
down. I can't think of a scenario where having the tarfile named with the
NOID offers any more protection.

So, how about we just call the directory "aip" and leave it at that? The
NOID then just has to be a string, with no responsibility for behaving well
in a file system.


Peter Binkley, Ph.D., MLIS / Digital Initiatives Technology Librarian /
peter.binkley@ualberta.ca
2-10K Cameron Library / University of Alberta / Edmonton, Alberta / Canada
T6G 2J8
phone 780-492-3743 / fax 780-492-9243

cwant:
I'm not an Swift expert, but don't you need a unique path to deposit to? If two preservation objects are called 'aip.tar', won't they clobber each other in Swift?

pbinkley:
The object name is optional in a Swift PUT, which I was thinking meant you
can let Swift make up its own. But I see that's for creating containers
rather than unnamed objects. So we're back to using the sanitized NOID, or
generating a UUID and storing it in the Fedora object for re-use in
preservation operations. We can do the UUID approach without having to
replace the NOID as the primary identifier in Fedora. Or we could sanitize
the NOIDs using Pairtree identifier string cleaning, as Fedora does, and
keep the storage-level identifiers within Fedora and Swift in alignment.


Peter Binkley, Ph.D., MLIS / Digital Initiatives Technology Librarian /
peter.binkley@ualberta.ca
2-10K Cameron Library / University of Alberta / Edmonton, Alberta / Canada
T6G 2J8
phone 780-492-3743 / fax 780-492-9243

PR #71, opened 2017-06-02
Make sure we are adding metadata to swift deposits as in AIP spec

murny:
Something we haven't been doing yet.

From according to AIP spec( https://docs.google.com/document/d/154BqhDPAdGW-I9enrqLpBYbhkF9exX9lV3kMaijuwPg/edit) we need to add metadata when we upload our deposited files in Swift:

```
Ingest to Swift
Swift container: “ERA”
Swift path: <NOID>
 
The identifier in Swift will normally be the NOID generated in ERA. It may be some other identifier for non-ERA content (an identifier used in that project, or a UUID if nothing else is available). Swift X-Object-Meta name-value pairs will be populated:
 
project: normally “ERA” (this may become redundant if we always use a container per project, but we haven’t determined that yet)
Project_id: (i.e. the id of this object within the project) - normally the NOID
promise: a label for the level of preservation promised 
This will ultimately cover various levels of preservation as defined in the UAL Digital Preservation Plan (gold, silver, bronze). For lightweight AIPS this will always be “bronze”.
retention: (optional) a timestamp indicating when some action expressed by the promise will be taken
Not applicable to ERA content, but for other content we might use it to trigger a review after 10 years etc.
depositor: (optional) name of a non-UAL depositor
AIPversion: 1.0
```

From looking at it, the metadata on every file in PushmiPullyu should look something like as a result:
```ruby
  metadata = {
    project: 'ERA',
    project_id: file_base_name, #we name the files with the noid so should be the same
    promise: 'bronze',
    aip_version: '1.0'
    }
```

### DIFF
--- a/lib/pushmi_pullyu/swift_depositer.rb
+++ b/lib/pushmi_pullyu/swift_depositer.rb
@@ -28,7 +28,24 @@ class PushmiPullyu::SwiftDepositer
                      else
                        era_container.create_object(file_base_name)
                      end
-    deposited_file.write(File.open(file_name), 'etag' => hash, 'content-type' => 'application/x-tar')
+
+    # Add swift metadata with in accordance to AIP spec:
+    # https://docs.google.com/document/d/154BqhDPAdGW-I9enrqLpBYbhkF9exX9lV3kMaijuwPg/edit#
+    metadata = {
+      project: 'ERA',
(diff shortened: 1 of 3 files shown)


murny:
Seems to be failing on ruby 2.3 but works on ruby 2.4. Investigating.

PR #74, opened 2017-06-07
Integration test

### LINKED ISSUE #27: Tests and Specs (corner cases and VCR setup)

murny:
closes #27

### DIFF
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -49,9 +49,6 @@ class PushmiPullyu::CLI
     setup_log
     print_banner
 
-    setup_queue
-    setup_swift
-
     run_tick_loop
   end
 
@@ -160,28 +157,31 @@ class PushmiPullyu::CLI
     PushmiPullyu.reset_logger = false
   end
 
+  def run_preservation_cycle
+    item = queue.wait_next_item
+
+    # add additional information about the error context to errors that occur while processing this item.
+    Rollbar.scoped(noid: item) do
+      begin
+        # Download AIP from Fedora, bag and tar AIP directory and cleanup after block code
+        PushmiPullyu::AIP.create(item) do |aip_filename|
+          # Push tarred AIP to swift API
+          deposited_file = swift.deposit_file(aip_filename, options[:swift][:container])
+          # Log successful preservation event to the log files
+          PushmiPullyu::Logging.log_preservation_event(deposited_file)
+        end
+      rescue => e
+        Rollbar.error(e)
+        logger.error(e)
+        # TODO: we could re-raise here and let the daemon die on any preservation error, or just log the issue and
+        # move on to the next item.
+      end
+    end
+  end
+
   def run_tick_loop
     while PushmiPullyu.server_running?
-      # Preservation (TODO):
-      item = @queue.wait_next_item
-
-      # add additional information about the error context to errors that occur while processing this item.
-      Rollbar.scoped(noid: item) do
-        begin
-          # Download AIP from Fedora, bag and tar AIP directory and cleanup after block code
-          PushmiPullyu::AIP.create(item) do |aip_filename|
-            # Push tarred AIP to swift API
-            deposited_file = @storage.deposit_file(aip_filename, options[:swift][:container])
-            # Log successful preservation event to the log files
-            PushmiPullyu::Logging.log_preservation_event(deposited_file)
-          end
-        rescue => e
-          Rollbar.error(e)
-          # TODO: we could re-raise here and let the daemon die on any preservation error, or just log the issue and
-          # move on to the next item.
-        end
-      end
-
+      run_preservation_cycle
       rotate_logs if PushmiPullyu.reset_logger?
     end
   end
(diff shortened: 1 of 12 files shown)


murny:
Reason why this PR is so huge is because of the VCR tape. Most of it is the data of the binary file (AIP tar) but also due to the fact that Swift openstack doesn't cache any of the calls for getting it's data/metadata. So every call to the deposited_file object (like getting metadata values, etag, modifed date, etc) do a HEAD request to the API.

Perhaps we can do this smarter(clone the object and pass that around or something instead of doing a bunch of API calls)

lib/pushmi_pullyu/cli.rb
```diff
@@ -160,28 +157,31 @@ def rotate_logs
...
-          # move on to the next item.
+  def run_preservation_cycle
+    item = queue.wait_next_item
+
+    # add additional information about the error context to errors that occur while processing this item.
+    Rollbar.scoped(noid: item) do
+      begin
+        # Download AIP from Fedora, bag and tar AIP directory and cleanup after block code
+        PushmiPullyu::AIP.create(item) do |aip_filename|
+          # Push tarred AIP to swift API
```
mbarnett: I wonder if it would be useful to nest another Rollbar scope for the filename in here?

spec/integration/acceptance_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  let(:aip_folder) { "#{workdir}/#{noid}" }
+  let(:aip_file) { "#{aip_folder}.tar" }
+  let(:log_folder) { "#{workdir}/log" }
+
+  let!(:opts) { PushmiPullyu.options.dup }
+  let!(:redis) { Redis.new }
+
+  before do
+    # setup redis and add noid to queue
+    redis.zadd 'test:pmpy_queue', 10, noid
```
mbarnett: you might want to clear the queue before the add? It's borderline pointless but I did it in my tests just to ensure if another test missed clearing it after, things would still work as expected.

spec/pushmi_pullyu/preservation_queue_spec.rb
```diff
@@ -4,15 +4,18 @@
 RSpec.describe PushmiPullyu::PreservationQueue do
   describe 'a queue with 3 items in it' do
     let(:queue) { PushmiPullyu::PreservationQueue.new(poll_interval: 0, queue_name: 'test:pmpy_queue') }
+    let!(:redis) { Redis.new }
```
mbarnett: Doesn't it yell at you about not using `let` variables in before blocks? That's why I avoided it.

spec/integration/acceptance_spec.rb
murny: I changed yours and am doing this in the after/teardown block. So you can hopefully assume redis is always clean to begin with. Instead of being parnoid. I ran the test suite like 50 times in random orders trying to catch weird things (and i did catch a few, reason why theres a few unreleated changed in this PR)

spec/pushmi_pullyu/preservation_queue_spec.rb
murny: I wanted to clean redis in the after block. So my options is to create an instance variable in the before block so that I can reuse in the after block (which rubocop rspec will yell at you for). Or just create a let banged and use that variable which is what I did instead. 
These `let`s are also before the before block also.

lib/pushmi_pullyu/cli.rb
```diff
@@ -160,28 +157,31 @@ def rotate_logs
...
+          deposited_file = swift.deposit_file(aip_filename, options[:swift][:container])
+          # Log successful preservation event to the log files
+          PushmiPullyu::Logging.log_preservation_event(deposited_file)
         end
+      rescue => e
+        Rollbar.error(e)
+        logger.error(e)
```
murny: I am logging the error now too (was swallowing and silently failing in my test suite)

lib/pushmi_pullyu/cli.rb
murny: What do these scopes look like in rollbar? Suppose it cant hurt. But in reality the filename is just the `noid` with a `.tar` extension which doesnt give too much more information as we are already scoping the `noid`. But it does give you the information that it succeed the AIP creation at least.

mbarnett: It's basically extra context added to the alert for any exception that's thrown within the scope. So anything that blows up in the noid scope gets tagged with the noid, etc.

cwant:
I still think you should at least through a NoidInvalid if the noid has a slash in it. (And then write a test for it.)

Sure, we don't have noids with slashes in them today (and likely not tomorrow), but since this is an open source project and a noid with a slash will break execution of this program (and worse, represents a security risk), it's probably worth doing.

mbarnett:
Yeah, I'd forgotten about that discussion, but it's a good idea

murny:
We now throw a exception if a noid contains a slash(s) in it. This is done

cwant:
:100:

PR #77, opened 2017-06-09
Chage code to write file content during file creation

astrilet:
We have small bug in the current swift_depositor.rb code. When new swift object is created we use two API calls - one to create an object, another to write data to it. When archiving is turned on on the container, that cases file archive of 0 size been created when data in the file being updated. Fix is simple, just need to write file data in the create method.

### DIFF
--- a/lib/pushmi_pullyu/swift_depositer.rb
+++ b/lib/pushmi_pullyu/swift_depositer.rb
@@ -23,15 +23,11 @@ class PushmiPullyu::SwiftDepositer
 
     era_container = swift_connection.container(swift_container)
 
-    deposited_file = if era_container.object_exists?(file_base_name)
-                       era_container.object(file_base_name)
-                     else
-                       era_container.create_object(file_base_name)
-                     end
+    object_exists = era_container.object_exists?(file_base_name)
 
     # Add swift metadata with in accordance to AIP spec:
     # https://docs.google.com/document/d/154BqhDPAdGW-I9enrqLpBYbhkF9exX9lV3kMaijuwPg/edit#
-    metadata = {
+    swift_metadata = {
       project: 'ERA',
       project_id: file_base_name,
       promise: 'bronze',
@@ -39,11 +35,18 @@ class PushmiPullyu::SwiftDepositer
     }
 
     # ruby-openstack wants all keys of the metadata to be named like "X-Object-Meta-{{Key}}", so update them
-    metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
+    swift_metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
 
-    deposited_file.write(File.open(file_name),
-                         { 'etag' => checksum,
-                           'content-type' => 'application/x-tar' }.merge(metadata))
+    # creatre file metadata
+    file_metadata = { 'etag' => checksum,
+                      'content-type' => 'application/x-tar' }.merge(swift_metadata)
+
+    # create new file
+    return era_container.create_object(file_base_name, file_metadata, File.open(file_name)) unless object_exists
+
+    # update existing one
+    deposited_file = era_container.object(file_base_name)
+    deposited_file.write(File.open(file_name), file_metadata)
 
     deposited_file
   end
--- a/lib/pushmi_pullyu/swift_depositer.rb
+++ b/lib/pushmi_pullyu/swift_depositer.rb
@@ -23,11 +23,9 @@ class PushmiPullyu::SwiftDepositer
 
     era_container = swift_connection.container(swift_container)
 
-    object_exists = era_container.object_exists?(file_base_name)
-
     # Add swift metadata with in accordance to AIP spec:
     # https://docs.google.com/document/d/154BqhDPAdGW-I9enrqLpBYbhkF9exX9lV3kMaijuwPg/edit#
-    swift_metadata = {
+    metadata = {
       project: 'ERA',
       project_id: file_base_name,
       promise: 'bronze',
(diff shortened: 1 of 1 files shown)


lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -23,27 +23,30 @@ def deposit_file(file_name, swift_container)
...
-                       era_container.object(file_base_name)
-                     else
-                       era_container.create_object(file_base_name)
-                     end
+    object_exists = era_container.object_exists?(file_base_name)
 
     # Add swift metadata with in accordance to AIP spec:
     # https://docs.google.com/document/d/154BqhDPAdGW-I9enrqLpBYbhkF9exX9lV3kMaijuwPg/edit#
-    metadata = {
+    swift_metadata = {
```
murny: Calling this swift_metadata is redundant. We know it's swift metadata as we are in the swift_depositer class and interacting with swift. I named it metadata as this is the same naming scheme used in Openstack Ruby gem: https://github.com/ruby-openstack/ruby-openstack/blob/master/lib/openstack/swift/storage_object.rb#L11

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -23,27 +23,30 @@ def deposit_file(file_name, swift_container)
...
 
     # ruby-openstack wants all keys of the metadata to be named like "X-Object-Meta-{{Key}}", so update them
-    metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
+    swift_metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
 
-    deposited_file.write(File.open(file_name),
-                         { 'etag' => checksum,
-                           'content-type' => 'application/x-tar' }.merge(metadata))
+    # creatre file metadata
+    file_metadata = { 'etag' => checksum,
```
murny: I would call this `headers` to remain consistent with Ruby Openstack gem (https://github.com/ruby-openstack/ruby-openstack/blob/master/lib/openstack/swift/storage_object.rb#L49) and also because most of this stuff doesn't have anything to do with the file's metadata.

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -23,27 +23,30 @@ def deposit_file(file_name, swift_container)
 
     era_container = swift_connection.container(swift_container)
 
-    deposited_file = if era_container.object_exists?(file_base_name)
-                       era_container.object(file_base_name)
-                     else
-                       era_container.create_object(file_base_name)
-                     end
+    object_exists = era_container.object_exists?(file_base_name)
```
murny: Assigned a variable but never used more then once. Do you need this? Does it improve code readability? Probably not needed then?

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -23,27 +23,30 @@ def deposit_file(file_name, swift_container)
...
 
-    deposited_file.write(File.open(file_name),
-                         { 'etag' => checksum,
-                           'content-type' => 'application/x-tar' }.merge(metadata))
+    # creatre file metadata
+    file_metadata = { 'etag' => checksum,
+                      'content-type' => 'application/x-tar' }.merge(swift_metadata)
+
+    # create new file
+    return era_container.create_object(file_base_name, file_metadata, File.open(file_name)) unless object_exists
```
murny: Using return makes this hard to follow as you are abrupting the code flow and returning early. Also if someone comes into this file and wonders what this returns and sees the deposited_file variable being returned at the bottom, they will make the wrong assumption that this is what gets returned on a new file upload. This could all be simplified as such:

```ruby

    if era_container.object_exists?(file_base_name)
      deposited_file = era_container.object(file_base_name)
      deposited_file.write(File.open(file_name), file_metadata)
    else
      deposited_file = era_container.create_object(file_base_name, file_metadata, File.open(file_name))
    end

    deposited_file
```
Therefore you are always returning the same thing and the code flows much better (instead of being interrupted early and returning before it hits the end block)

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -23,27 +23,30 @@ def deposit_file(file_name, swift_container)
...
     }
 
     # ruby-openstack wants all keys of the metadata to be named like "X-Object-Meta-{{Key}}", so update them
-    metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
+    swift_metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
 
-    deposited_file.write(File.open(file_name),
-                         { 'etag' => checksum,
-                           'content-type' => 'application/x-tar' }.merge(metadata))
+    # creatre file metadata
```
murny: Majority of these little code comments aren't needed. Your code should be self explanatory. Telling me `era_container.create_object` creates a new file or creating a hash for file_metadata and telling me `create file metdata` is not needed as the code tells you this and is self explanatory.

Also you spelt `create` wrong here.

lib/pushmi_pullyu/swift_depositer.rb
astrilet: OK will change that back

lib/pushmi_pullyu/swift_depositer.rb
astrilet: good idea, will fix

lib/pushmi_pullyu/swift_depositer.rb
astrilet: will modify code as suggested

lib/pushmi_pullyu/swift_depositer.rb
astrilet: will fix as suggested

PR #90, opened 2017-08-29
#89 Update redis config to accept URL instead of host/port as you may…

### LINKED ISSUE #89: Update PushMiPullYu Redis connection to use url
This is related to https://github.com/ualbertalib/pushmi_pullyu/issues/87

Current configuration of PMPY uses host/port for redis connection. https://github.com/ualbertalib/pushmi_pullyu/blob/master/lib/pushmi_pullyu/cli.rb#L205-L211

This should be updated to use redis_url as _redis:// URL_:

```redis://:p4ssw0rd@10.0.1.1:6380/15```

So that it can use a password for authentication, and potentially different database as needed.

murny:
… need further configuration like password for connecting to redis

closes #89

### DIFF
--- a/examples/pushmi_pullyu.yml
+++ b/examples/pushmi_pullyu.yml
@@ -19,8 +19,7 @@ minimum_age: 0
 rollbar_token: 'abc123xyz'
 
 redis:
-  host: localhost
-  port: 6379
+  url: redis://localhost:6379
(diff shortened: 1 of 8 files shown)


weiweishi:
LGTM. Thanks for working on this.

PR #98, opened 2017-09-25
Fixes for keystone v3 authorization

astrilet:
Changes required for keystone authorization v3 to work

### DIFF
--- a/lib/pushmi_pullyu.rb
+++ b/lib/pushmi_pullyu.rb
@@ -43,11 +43,12 @@ module PushmiPullyu
       base_path: '/dev'
     },
     swift: {
-      auth_version: 'v1.0',
       tenant: 'tester',
       username: 'test:tester',
       password: 'testing',
-      endpoint: 'http://localhost:8080',
+      auth_url: 'http://localhost:8080/auth/v1.0',
+      project_name: 'demo',
+      project_domain_name: 'default',
       container: 'ERA'
     }
   }.freeze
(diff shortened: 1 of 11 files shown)


lib/pushmi_pullyu.rb
```diff
@@ -43,11 +43,12 @@ module PushmiPullyu
       base_path: '/dev'
     },
     swift: {
-      auth_version: 'v1.0',
       tenant: 'tester',
       username: 'test:tester',
       password: 'testing',
-      endpoint: 'http://localhost:8080',
+      auth_url: 'http://localhost:8080/auth/v1.0',
```
weiweishi: Since we are not using v1.0, I think the example used in these configurations should set up to as close to our true configuration as possible to avoid confusion?

weiweishi: Never mind. This is used to facilitate testing which builds in v1.0? Is it possible to upgrade our swift testing env to use the latest version, so it is actually useful?

astrilet: our docker image uses v1.0 authentication (config files authentication vs keystone). This is why it is v1.0 keystone would require a bit to work to be installed inside docker image.

lib/pushmi_pullyu/cli.rb
```diff
@@ -170,7 +170,7 @@ def run_preservation_cycle
           # Log successful preservation event to the log files
           PushmiPullyu::Logging.log_preservation_event(deposited_file)
         end
-      rescue => e
+      rescue StandardError => e
```
murny: I would have this one how it was. We want to capture all errors. You can disable this in RoboCop for this one occurrence. By using an inline comment, something like `# rubocop:disable <name of cop error>`

astrilet: OK, this is good point, will change that

astrilet: All changes have been done. Please review again.

PR #104, opened 2017-10-05
Fix missing metadata in swift

### LINKED ISSUE #105: Missing Metadata in new object creation
During the integration tests with ERA-Test and PMPY, we notice that the metadata for OpenStack swift object is missing when a new object is created, and only when the object is updated, the metadata will be posted to OpenStack. We identify this as [the bug in the upstread ruby-openstack gem](https://github.com/ruby-openstack/ruby-openstack/issues/94):

Header hash expected by [create method](https://github.com/ruby-openstack/ruby-openstack/blob/master/lib/openstack/swift/storage_object.rb#L49-L63):
{ etag: 'some_value',
content_type: 'application/x-tar',
metadata: { "X-Object-Meta-project"=>"MyContainer",
"X-Object-Meta-project_id"=>"testABC",
...}
}
this method is used when depositing new object into object storage.

Header hash expected by [write method](https://github.com/ruby-openstack/ruby-openstack/blob/master/lib/openstack/swift/storage_object.rb#L255-L265) when updating existing object:

{"etag"=>"22affe973e2175a72a11c9ecb1acc664",
"content-type"=>"application/x-tar",
"X-Object-Meta-project"=>"MyContainer",
"X-Object-Meta-project_id"=>"testABC",
....}

astrilet:
when creating new file in swift metadata was missing. fix #105

### DIFF
--- a/lib/pushmi_pullyu/swift_depositer.rb
+++ b/lib/pushmi_pullyu/swift_depositer.rb
@@ -37,7 +37,9 @@ class PushmiPullyu::SwiftDepositer
     # ruby-openstack wants all keys of the metadata to be named like "X-Object-Meta-{{Key}}", so update them
     metadata.transform_keys! { |key| "X-Object-Meta-#{key}" }
 
+    # headers hash need to be different for write vs create_object methods
     if era_container.object_exists?(file_base_name)
+      # write method expects all key/value pairs to be strings
       headers = { 'etag' => checksum,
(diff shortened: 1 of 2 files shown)


astrilet:
needed to change how new objects are created in swift, if we create it with call 
create_object(file_base_name, headers, File.open(file_name))
metadata was not being updated.

PR #111, opened 2017-10-27
update rollbar config to allow it to use proxy

weiweishi:
In order to use rollbar through a proxy, need to let PMPY to talk to the endpoint on the proxy.


murny:
Is endpoint what we want?

They do provide a proxy option for...web proxies?

Perhaps this might be better suited?
https://github.com/rollbar/rollbar-gem#web-proxies

weiweishi:
Henry is going to use https_proxy mentioned in their documentation but we are waiting for https://github.com/rollbar/rollbar-gem to release a new version of the gem so that we can use the proxy configuration. Closing this PR, and will open a new one once new rollbar gem is in place.

PR #119, opened 2017-11-03
to use proxy for rollbar monitoring

weiweishi:
To configure rollbar to use proxy.


examples/pushmi_pullyu.yml
```diff
@@ -17,6 +17,8 @@ process_name: pushmi_pullyu
 queue_name: dev:pmpy_queue
 minimum_age: 0
 rollbar_token: 'abc123xyz'
+rollbar_proxy_host: 'your_proxy_host_url'
```
murny: Wondering if it be better to name space rollbar now? like the rest?

```rollbar: 
  token: 'etc'
  proxy_host: 'etc'
  proxy_url: 'etc
```

Also might not hurt to write a test to make sure these are being set accordingly? (copy and paste the `rollbar_token` test if you want)

lib/pushmi_pullyu/cli.rb
```diff
@@ -58,6 +58,10 @@ def configure_rollbar
     Rollbar.configure do |config|
       config.enabled = false unless options[:rollbar_token].present?
       config.access_token = options[:rollbar_token]
+      config.proxy = {
+        host: options[:rollbar_proxy_host],
```
murny: assuming rollbar still works if these are nils and just doesnt use proxy?

lib/pushmi_pullyu/cli.rb
```diff
@@ -225,7 +229,7 @@ def shutdown
       exit!(1)
     else
       # using stderr instead of logger as it uses an underlying mutex which is not allowed inside trap contexts.
-      $stderr.puts 'Exiting...  Interrupt again to force quit.'
+      warn 'Exiting...  Interrupt again to force quit.'
```
murny: theres a test that makes this fail. Requiring updating it. Look at travisCI build

weiweishi:
Okay - will do.

One thing comes up in fixing a rubocop error here:
https://github.com/ualbertalib/pushmi_pullyu/pull/119/files/9fb8a729f0a4e1c42ef9960f8e845a6c63e708b3#diff-8d1e664a20e3c2b30fe4f01a34477acdL228

This breaks a spec test: /spec/pushmi_pullyu/cli_spec.rb:71

Not sure how to fix it. still reading on it.


*Weiwei Shi*

Digital Initiative Applications Librarian
University of Alberta Libraries
2-10L Cameron Library
Edmonton, Alberta, Canada T6G 2J8
Phone:(780)492-7802
Fax: (780)248-1209
Email: weiwei.shi@ualberta.ca

lib/pushmi_pullyu/cli.rb
weiweishi: the spec test still passes, without the two values set up in fixtures/config.yml, so it should be working without the proxy values

lib/pushmi_pullyu/cli.rb
murny: Take a peak at changing the test to using these maybe?

https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/output-matcher

lib/pushmi_pullyu/cli.rb
murny: Well the spec test literally just checks if the config file gets read and initializes the option hash. Doesn't actually test if rollbar gets configured. Haven't read into rollbar. I'm assuming it works (if it does this is fine)? Or maybe it doesn't (then we need to do a present check or something) ? ¯\\_(ツ)_/¯

weiweishi:
thanks @murny

PR #139, opened 2018-03-01
Making PMPY defaults match the Jupiter defaults for fedora.

cwant:
Also: although mostly cosmetic, making the tests refer to 'test' rather than 'dev'.


spec/pushmi_pullyu/aip/creator_spec.rb
```diff
@@ -80,6 +80,9 @@
          '50b065c7cf19ed3e282a2a98b70f6e9429cc56ea data/objects/metadata/content_fcr_metadata.n3',
          '5eb6d58841f4196cc682ef1af3054dddacb6d40c data/objects/whatever.pdf'].sort
 
+      puts lines
```
murny: Was this just for debugging? Should it be left here?

cwant: Yeah, totally debugging -- thanks for catching that.

PR #148, opened 2018-07-25
updated version number for new tagged release

weiweishi:
Overlooked the version change when first tagging the release. 
Updated the version number. Will tag a new release 1.0.1 with this change.

### DIFF
--- a/lib/pushmi_pullyu/aip.rb
+++ b/lib/pushmi_pullyu/aip.rb
@@ -17,5 +17,7 @@ module PushmiPullyu::AIP
     FileUtils.rm_rf(aip_filename) if File.exist?(aip_filename)
     FileUtils.rm_rf(aip_directory) if File.exist?(aip_directory)
   end
+
+  # rubocop:disable Style/AccessModifierDeclarations
   module_function :create
 end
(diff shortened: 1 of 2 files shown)


murny:
Looks fine, however build is failing due to rubocop jumping a version.

So can either, lock rubocop to a certain version or fix the error. Error is a bit confusing but from looking at rubocop issues, https://github.com/rubocop-hq/rubocop/issues/5953 and https://github.com/rubocop-hq/rubocop/issues/6032, you should be able to simply just move the `module_function` line of code before the create method, so it acts similarilary to a `private/protected` method.

https://github.com/ualbertalib/pushmi_pullyu/blob/master/lib/pushmi_pullyu/aip.rb

Change:
```ruby
module PushmiPullyu::AIP
  def create(noid)
    # Stuff goes here
  end

  module_function :create
end
```
To this:
```ruby
module PushmiPullyu::AIP
  module_function

  def create(noid)
    # Stuff goes here
  end
end
```

If not, then just inline a disable rubocop on the offending line
`# rubocop:disable Style/AccessModifierDeclarations`

weiweishi:
yep. it worked. thanks for pointing it out. it needs a reapprove now @murny

weiweishi:
image

PR #149, opened 2018-08-17
Add authentication to allow list source to be loaded

weiweishi:
To allow the RDF gem to load the graph so that it can get the ordering of the files, it needs to authorize the access to list_source.

### DIFF
--- a/lib/pushmi_pullyu/aip/file_list_creator.rb
+++ b/lib/pushmi_pullyu/aip/file_list_creator.rb
@@ -1,5 +1,6 @@
 require 'rdf'
 require 'rdf/n3'
+require 'rest-client'
 
 class PushmiPullyu::AIP::FileListCreator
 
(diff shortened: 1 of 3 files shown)


lib/pushmi_pullyu/aip/file_list_creator.rb
```diff
@@ -20,6 +21,9 @@ class ListSourceFileSetMismatch < StandardError; end
 
   def initialize(list_source_uri, output_xml_file, file_set_uuids)
     @uri = RDF::URI(list_source_uri)
+    @auth_uri = RDF::URI(list_source_uri)
+    @auth_uri.user = PushmiPullyu.options[:fedora][:user]
```
weiweishi: @uri is also used as the subject for the triples, so it can't contain the user/pass in the uri. It needs a auth_uri to just allow the graph to load.

pushmi_pullyu.gemspec
```diff
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'rdf', '>= 1.99', '< 4.0'
   spec.add_runtime_dependency 'rdf-n3', '>= 1.99', '< 4.0'
   spec.add_runtime_dependency 'redis', '>= 3.3', '< 5.0'
+  spec.add_runtime_dependency 'rest-client', '~> 1.8'
```
weiweishi: This is the dependency that allows the rdf gem to authenticate with username/password.

weiweishi:
Thanks @murny! Totally didn't expect you'd be checking email, and working
at this hour! much appreciated!



*Weiwei Shi*Digital Initiative Applications Librarian

2-10L Cameron Library, University of Alberta
780-492-7802  |  weiwei.shi@ualberta.ca
*"The University of Alberta respectfully acknowledges that we are situated
on Treaty 6 territory, traditional lands of First Nations and Métis
people."*

PR #151, opened 2018-08-22
fixed the db connection, remove false positive check on existing db c…

weiweishi:
…onnection

### DIFF
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -268,17 +268,7 @@ class PushmiPullyu::CLI
   end
 
   def database_configuration
-    # Config either from URL, or with more granular options (the later taking precedence)
-    config = {}
-    uri = URI.parse(PushmiPullyu.options[:database][:url])
-    config[:adapter] = PushmiPullyu.options[:database][:adaptor] || uri.scheme
-    config[:host] = PushmiPullyu.options[:database][:host] || uri.host
-    config[:database] = PushmiPullyu.options[:database][:database] || uri.path.split('/')[1].to_s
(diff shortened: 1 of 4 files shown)


murny:
Rubocop takes precedence over hound. So as long as rubocop is happy that's all that matters. This typically happens because Hound just follows it's own version of rubocop (no way of locking it to one), so sometimes its either on an older/newer version than rubocop. 

(Also Hound is no longer a required check either).

weiweishi:
Okay! Good to know!

lib/pushmi_pullyu/cli.rb
```diff
@@ -262,4 +263,22 @@ def start_server_as_daemon
     end
   end
 
+  def setup_db_connection
+    ActiveRecord::Base.establish_connection(database_configuration)
+  end
+
+  def database_configuration
```
murny: Not a huge problem, but what's your thoughts about simplifying this?

Can't see any reason why we need such granularity. Also I believe activerecord already does all this parsing and precedence on it's end already: https://edgeguides.rubyonrails.org/configuring.html#connection-preference 

This could just simply be passing in a url (we dont need pool/encoding/database_name/etc) or if we do want this granularity, I believe you can just pass in the database options hash (PushmiPullyu.options[:database]):
```ruby
def database_configuration
 PushmiPullyu.options[:database][:url]
end
```
This would also make configuration easier for henry/system admins as they can just copy the same `url:` from jupiter to pmpy. Instead of currently they being a little bit different for both applications

weiweishi: Yes, I agreed. I moved the code from previous method directly but we should take the chance to simplify it. 

Will make the change.

PR #175, opened 2018-11-29
modifying the filtering tranformation to filter out unwanted info

weiweishi:
* Made changes to Matt's original scrubbing method, based on the json structure of the rollbar payload:
      "exception": {
        "message": "<http://****:******@mycombe.library.ualberta.ca:8080/fedora/rest/prod/8c/ff/bb/0f/8cffbb0f-8cb8-4061-b6fe-3122a595d181/list_source>: 404", 
        "class": "IOError"
      }
* Added level: ignore for IOError messages. 

This should be considered as a WIP PR, as I don't have a good way to test the code changes, and I'd like to hear from others, which seems to be a better/more reliable solution, to avoid having our fedora password get into Rollbar again.


lib/pushmi_pullyu/cli.rb
```diff
@@ -59,13 +59,15 @@ def configure_rollbar
...
       config.access_token = options[:rollbar][:token]
-
+      config.exception_level_filters.merge!({
+       'IOError' => 'ignore'
+      })
       # add a filter after Rollbar has built the error payload but before it is delivered to the API,
       # in order to strip sensitive information out of certain error messages
       exception_message_transformer = proc do |payload|
-        clean_message = payload[:exception].message.sub(/http:\/\/.+:.+@(.+)\/fedora\/rest\/prod\/(.*)/,
+        clean_message = payload[:exception][:message].sub(/http:\/\/.+:.+@(.+)\/fedora\/rest\/prod\/(.*)/,
```
weiweishi: According to Rollbar, the payload json structured like this:
      "exception": {
        "message": "<http://****:*****r@mycombe.library.ualberta.ca:8080/fedora/rest/prod/8c/ff/bb/0f/8cffbb0f-8cb8-4061-b6fe-3122a595d181/list_source>: 404", 
        "class": "IOError"
      }
So we will need to replace the value in payload[:exception][:message] with the scrubbed version.

lib/pushmi_pullyu/cli.rb
```diff
@@ -59,13 +59,15 @@ def configure_rollbar
     Rollbar.configure do |config|
       config.enabled = false unless options[:rollbar][:token].present?
       config.access_token = options[:rollbar][:token]
-
+      config.exception_level_filters.merge!({
```
weiweishi: Add IOError as to be ignored so this particular error message will not be sent to Rollbar.

lib/pushmi_pullyu/cli.rb
```diff
@@ -38,7 +38,7 @@ def run
       end
     # rubocop:disable Lint/RescueException
     rescue Exception => e
-      Rollbar.error(e)
+      Rollbar.error(e, use_exception_level_filters: true)
```
murny: We can probably make this global? Not sure how many `Rollbar.error` calls we doing in pushmi_pullyu (might just be this one, but if someone ever decides to come in and add another one, they don't have to forgot adding this `use_exception_level_filters: true` call).

Can add it right to the config:

```
Rollbar.configure do |config|
  config.use_exception_level_filters = true
end
```

(oddly this is not in the documentation, but heres a PR for the commit: https://github.com/rollbar/rollbar-gem/pull/588)

lib/pushmi_pullyu/cli.rb
```diff
@@ -59,13 +59,13 @@ def configure_rollbar
     Rollbar.configure do |config|
       config.enabled = false unless options[:rollbar][:token].present?
       config.access_token = options[:rollbar][:token]
-
+      config.exception_level_filters['IOError'] = 'ignore'
```
murny: Maybe use their provided syntax? If we decide to filter another exception in the future, someone doesn't just come in and copy and paste this line, they can add it to the merge call:

```
config.exception_level_filters.merge!({
  'IOError' => 'ignore'
})
```

lib/pushmi_pullyu/cli.rb
murny: Ah the PR example is wrong?

Correct usage is `use_exception_level_filters_default = true`

https://github.com/rollbar/rollbar-gem/blob/master/docs/configuration.md#use_exception_level_filters_default

lib/pushmi_pullyu/cli.rb
```diff
@@ -59,13 +59,14 @@ def configure_rollbar
     Rollbar.configure do |config|
       config.enabled = false unless options[:rollbar][:token].present?
       config.access_token = options[:rollbar][:token]
-
+      config.use_exception_level_filters = true
```
murny:
```suggestion
      config.use_exception_level_filters_default = true
```

PR #176, opened 2018-11-29
regression for filename too long is fixed by minitar 0.7

### LINKED ISSUE #171: [PMPY] Handle long file name to avoid Archive::Tar::Minitar::FileNameTooLong
If the `tar_aip_filename` is  longer than 256 bytes here https://github.com/ualbertalib/pushmi_pullyu/blob/master/lib/pushmi_pullyu/aip/creator.rb#L44, minitar will throw an error message. 

Need a way to handle and potentially shorten the aip's filename. 

@pbinkley @umarqasim @sfarnel any suggestion if we need to shorten the filename for the aip package?

pgwillia:
fixes #171


lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -50,7 +50,7 @@ def deposit_file(file_name, swift_container)
     else
       # for creating new: construct hash with symbols as keys, add metadata as a hash within the header hash
       headers = { etag: checksum,
-                  content_type:  'application/x-tar',
+                  content_type: 'application/x-tar',
```
pgwillia: aligned for rubocop

spec/pushmi_pullyu/aip/creator_spec.rb
```diff
@@ -87,11 +88,12 @@
          '30604fd556c96aac01abddf1e8b4e0369fc90ba5 '\
          'data/objects/metadata/files_metadata/856444b6-8dd5-4dfa-857d-435e354a2ead/original_file_metadata.n3',
          '9ea739a91eff6ba99e0227e3a909436d1dfd7ca7 '\
-         'data/objects/files/01bb1b09-974d-478b-8826-2c606a447606/theses.jpg',
+         'data/objects/files/01bb1b09-974d-478b-8826-2c606a447606/Archive_Tar_Minitar_FileNameTooLongHjMQh5EggUDWgpsfoXfyNGoEbvcUcN34YPpjknJhu4y8bs3qVMvBXA5A2aFBxY6EiIyxRS.jpg',
```
murny: Instead of turning rubocop off, I would just break up this really long string. You can use the backslash to accomplish this.

The backslash, the line-continuation character, to tell Ruby that these are in fact the same statement, spread over multiple lines.

Think of it kinda like concatenation (long_string becomes short_string + short_string), except theirs no calculation/performance hit, as ruby treats these as the exact same thing.

```suggestion
         'data/objects/files/01bb1b09-974d-478b-8826-2c606a447606/'\
         'Archive_Tar_Minitar_FileNameTooLongHjMQh5EggUDWgpsfoXfyNGo'\
         'EbvcUcN34YPpjknJhu4y8bs3qVMvBXA5A2aFBxY6EiIyxRS.jpg',
```

PR #197, opened 2020-03-09
Add http authentication for Jupiter calls

lagoan:
Jupiter's AIP API is protected by username/password pairs; pushmi_pullyu
needs to provide identification credentials to be able to access
restricted data.

### DIFF
--- a/lib/pushmi_pullyu/aip/downloader.rb
+++ b/lib/pushmi_pullyu/aip/downloader.rb
@@ -4,6 +4,8 @@ require 'rdf'
 require 'rdf/n3'
 require 'net/http'
 require 'digest'
+require 'faraday'
+require 'faraday-cookie_jar'
 
 # Download all of the metadata/datastreams and associated data related to an object
 class PushmiPullyu::AIP::Downloader
@@ -18,6 +20,7 @@ class PushmiPullyu::AIP::Downloader
 
   class JupiterDownloadError < StandardError; end
   class JupiterCopyError < StandardError; end
+  class JupiterAuthenticationError < StandardError; end
 
   def initialize(entity, aip_directory)
     @entity = entity
@@ -27,6 +30,8 @@ class PushmiPullyu::AIP::Downloader
 
   def run
     PushmiPullyu.logger.info("#{@entity_identifier}: Retreiving data from Jupiter ...")
+    # binding.pry
+    authenticate_http_calls
     make_directories
 
     # Main object metadata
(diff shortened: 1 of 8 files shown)


lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -76,24 +81,30 @@ def compare_md5(local, remote_checksum)
...
   end
 
-  def authenticate_and_request(url)
-    uri = URI(url)
-    request = Net::HTTP::Get.new(uri)
-    # TODO: This basic_auth call is just a placeholder to be replaced when a proper authentication mechanism is setup on
-    # jupiter https://github.com/ualbertalib/jupiter/pull/1370#issuecomment-561799351
-    request.basic_auth('admin', 'admin@ualberta.ca')
+  def authenticate_http_calls
+    @connection = Faraday.new(
```
murny: Is there any reason why we need to bring in a HTTP library like faraday over standard lib's Net::HTTP? Is there something we cannot do with Net:HTTP?

I'm a big advocate of trying to slim down our dependencies where we can:
More info: https://www.mikeperham.com/2016/02/09/kill-your-dependencies/

spec/support/http_cache/vcr/aip_download_and_swift_upload.yml
```diff
@@ -1,22 +1,76 @@
 ---
 http_interactions:
+- request:
+    method: post
+    uri: http://localhost:3000/auth/system?email=ditech@ualberta.ca&password=correct%20horse%20battery%20staple
```
murny: I feel like password/email should be within the post body, not as query params 🤔 . This could be a potential security issue (rollbar for example, if this errored out, rollbar will expose this url on its dashboard and elsewhere which now includes the systems username/password)

lib/pushmi_pullyu/aip/downloader.rb
mbarnett: Yeah, I'll just add to this that I haven't had great experiences with Faraday; it's a more complicated wrapper around the built-in library, which I've used pretty extensively and have always found reliable.

One of the values we've really been trying to stick to as a team is that where there's a "standard" solution for Ruby or Rails, we should always prefer that over some third-party solution. This really helps with us having a small team and a lot of projects on the go; it's easier for everyone to learn the one standard tool than to learn Faraday for one codebase, HTTParty for another, rest-client for a third, etc, etc. This was a problem in the past with a lot of the Hydra projects, which used an inconsistent mix of faraday, rest-client, http.rb, and possibly some others I'm forgetting.

If there's anything I can help you figure out how to do with Net::HTTP, let me know; like I said, I've used it a lot. On the flip side, if we try our best with Net::HTTP but become absolutely convinced it isn't the right solution, we should do a thorough survey of our options and try to pick _one_ solution that we can standardize everywhere to avoid the inconsistency we saw with the Hydra projects.

lagoan: Faraday helps a lot to keep clean code for managing cookies after signing into the system. It works as a client and I did not want to reinvent the wheel. 

One can sign in with
```
response = @connection.post('auth/system') do |request|
	...
end
```

And from then on out all the calls use the same cookie simplifying things; easy peasy.

Will be happy to look at other suggestions you have.

spec/support/http_cache/vcr/aip_download_and_swift_upload.yml
lagoan: You are right about where to send the data! This is a post call so I figured the data I was sending would have been sent as part of the body. Will need to double check how the request is being made.

lagoan:
yay!

PR #198, opened 2020-04-06
Bump ruby version to 2.4.0

lagoan:
Update minimum required version and add configuration examples


pgwillia:
Is it possible to go further?  Ruby 2.4 end of life was the end of March 2020.

PR #199, opened 2020-04-07
Update ruby version to 2.5

lagoan:
Ruby 2.4 has EOL date of 2020-03-31 so we need to update the minimum
requirements.


lagoan:
I just started a conversation with Neil about this. I believe we are on 2.4 at the moment.

PR #206, opened 2020-08-10
Add github workflow

lagoan:
## Context

Adding github workflow brings the project in line with Jupiter following the same testing mechanism.

## What's New

This PR includes files describing the github workflow for CI, a template for PR descriptions, and definition of code owners.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -1,10 +1,3 @@
-# This workflow uses actions that are not certified by GitHub.
-# They are provided by a third-party and are governed by
-# separate terms of service, privacy policy, and support
-# documentation.
-# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
-# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
-
 name: CI
 
 on:
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -23,5 +23,9 @@ jobs:
       run: bundle install
     - name: Lint with RuboCop
       run: bundle exec rubocop --parallel
+    - name: Run Danger
+      env:
+        DANGER_GITHUB_API_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
+      run: bundle exec danger
     - name: Run tests
       run: bundle exec rake spec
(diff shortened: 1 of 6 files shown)


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the docker-compose stack
+      run: docker-compose -f docker-compose-ci.yml up -d
```
murny, line 22: if all we need is redis, then lets use github actions service? No need for this docker stuff. 

The reason you see this in Jupiter is we have a very weird dependency with Solr. We need our code checkout, which we push config into solr, but if we use services, we don't have access to our code. Basically chicken and egg problem.

Here you can just use this:

```
jobs:
  test:
    runs-on: ubuntu-latest

    services:
      redis:
        image: redis:alpine
        ports: ["6379:6379"]
 steps:
    - uses: actions/checkout@v2
 ....etc...
```

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will download a prebuilt Ruby version, install dependencies and run tests with Rake
+# For more information see: https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby
```
murny, line 6: Do we need this disclaimer?

.github/workflows/ruby.yml
lagoan: Thanks! This is definitely and improvement.

.github/workflows/ruby.yml
lagoan: No we dont

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.6.6
+    - name: Install dependencies
+      run: bundle install
+    - name: Lint with RuboCop
+      run: bundle exec rubocop --parallel
+    - name: Run tests
+      run: bundle exec rake spec
```
murny, line 27: Should we add Danger?

lagoan: We should indeed!

lagoan: You are right about the travisCI stuff. Will be running the checks in parallel for a bit, after we make sure everything is OK will be sending another PR removing travisCI.

PR #207, opened 2020-08-10
Fix danger error message

lagoan:
## Context

Rubocop complains about using fail in Dangerfile and insists on using raise instead. However, danger uses fail with the sticky flag to specify how to treat the error. See https://danger.systems/reference.html

## What's New

Have rubocop ignore Dangerfile

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -16,9 +16,8 @@ warn('Please add a detailed summary in the description.') if github.pr_body.leng
 # Let people say that this isn't worth a CHANGELOG entry in the PR if they choose
 declared_trivial = (github.pr_title + github.pr_body).include?('#trivial') || !has_app_changes
 
-if !git.modified_files.include?('CHANGELOG.md') && !declared_trivial
-  fail("Please include a CHANGELOG entry. \nYou can find it at " \
-    '[CHANGELOG.md](https://github.com/ualbertalib/jupiter/blob/master/CHANGELOG.md).', sticky: false)
+if !git.modified_files.include?("CHANGELOG.md") && !declared_trivial
+  fail("Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/ualbertalib/pushmi_pullyu/blob/master/CHANGELOG.md).", sticky: false)
 end
 
 # Warn when there is a big PR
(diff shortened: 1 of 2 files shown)


ConnorSheremeta:
Why make changes to the dangerfile if it is now excluded from rubocop?

lagoan:
Good question @ConnorSheremeta! Made me rethink of the problem.

If I understand correctly ```fail``` comes from danger and not the standard fail method. The ```raise``` method does not have the same parameter signature as dangers ```fail``` method. So rubocop wants to use ```raise``` on Dangerfile but that does not work due to the parameter mismatch

So in order for the tests to pass, we ignore rubocop and use dangers ```fail```.

Hope that makes sense and its correct.

Dangerfile
```diff
@@ -17,8 +17,9 @@ warn('Please add a detailed summary in the description.') if github.pr_body.leng
 declared_trivial = (github.pr_title + github.pr_body).include?('#trivial') || !has_app_changes
 
 if !git.modified_files.include?('CHANGELOG.md') && !declared_trivial
-  raise("Please include a CHANGELOG entry. \nYou can find it at " \
-    '[CHANGELOG.md](https://github.com/ualbertalib/jupiter/blob/master/CHANGELOG.md).', sticky: false)
+  error_message = "Please include a CHANGELOG entry. \nYou can find it at " \
+                  '[CHANGELOG.md](https://github.com/ualbertalib/pushmi_pullyu/blob/master/CHANGELOG.md).'
+  fail(error_message, sticky: false)
```
murny, line 22: Minor, but variable assignment isn't really needed? Can just stick the string here and style this code accordingly?

lagoan: I am using variable assignment to keep things tidy and readable. If I use the full string in a single line we go over the 120 character limit:
```
  fail("Please include a CHANGELOG entry. \nYou can find it at [CHANGELOG.md](https://github.com/ualbertalib/pushmi_pullyu/blob/master/CHANGELOG.md)", sticky: false)
```
And separating the string with ```\``` inside the call makes it less readable:

```

  fail("Please include a CHANGELOG entry. \nYou can find it at " \
       '[CHANGELOG.md](https://github.com/ualbertalib/pushmi_pullyu/blob/master/CHANGELOG.md).', sticky: false)
```

PR #220, opened 2021-02-02
Fix dependency

mbarnett:
## Context

Henry noticed when trying to take this to prod that it won't boot because we `require 'uuid'` for core functionality but accidentally declared it as a development dependency only rather than a runtime dep.

## What's New

Change UUID to a runtime dependency, and bumped the version to 2.0.1 since we already pushed 2.0.0 to rubygems with the bad spec.


pgwillia (approved):
👍

Issue #200, opened 2020-04-14
Remove pg gem requirement


lagoan:
This gem should be removed because a direct connection to a database is no longer needed.

PR #241, opened 2021-11-22
Remove pg dependency

pgwillia:
## Context

This gem should be removed because a direct connection to a database is no longer needed. 

Related to #200 

## What's New

Removed gem dependency


jefferya (approved):

murny:
Can we remove the configuration stuff around the database (in config and examples yml): 
https://github.com/ualbertalib/pushmi_pullyu/search?q=database

Otherwise think this is good. I assume we pull all the data we require from the API and files themselves from the file system? Looks like nothing is using this database stuff anymore 👍

murny (approved):

PR #250, opened 2022-04-20
Bump ruby

pgwillia:
## Context

Ruby 2.6 was end of life March 2022.

## What's New

Ruby 2.7 ready.  

Added the Gemfile.lock after discussion on Dev Edmonton Slack.  
> As a result, you SHOULD check your Gemfile.lock into version control, in both applications and gems. If you do not, every machine that checks out your repository (including your production server) will resolve all dependencies again, which will result in different versions of third-party code being used if any of the gems in the [Gemfile(5)](https://bundler.io/man/gemfile.5.html) or any of their dependencies have been updated.
https://bundler.io/man/bundle-install.1.html#THE-GEMFILE-LOCK


.github/workflows/ruby.yml
```diff
@@ -18,9 +18,8 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6.6
-    - name: Install dependencies
-      run: bundle install
+        ruby-version: 2.7.6 # blocked from 3.0 see https://github.com/ualbertalib/pushmi_pullyu/issues/251
+        bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - name: Lint with RuboCop
```
pgwillia, line 23: I was getting an error in CI that I couldn't reproduce in development.  Turns out that the rubocop versions were different because we don't check in the Gemfile.lock.  We don't check in Gemfile.lock for ruby gems because that would make dependencies more rigid than we would want them to be.

henryzhang87 (commented):
I reviewed a few changes. Only question is the target ruby version changed from 2.5 to 2.6.

I am just curious why not 2.7 (what we are using). I am not in the capacity of "Approving" due to my limited knowledge of Ruby.

.rubocop.yml
```diff
@@ -12,7 +12,7 @@ AllCops:
     - 'vendor/**/*'
     - 'Dangerfile'
   ExtraDetails: true
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.6
```
ConnorSheremeta, line 15: Should this be 2.7? When we went to 2.5 we had it at 2.5
 https://github.com/ualbertalib/pushmi_pullyu/commit/99b1d8f8882cdfcef153bbd19a5230b5ec969a4d

pushmi_pullyu.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.5'
+  spec.required_ruby_version = '>= 2.6'
```
ConnorSheremeta, line 22: Again, should this be 2.7? When we went to 2.5 we had it at 2.5
 https://github.com/ualbertalib/pushmi_pullyu/commit/99b1d8f8882cdfcef153bbd19a5230b5ec969a4d

pgwillia: We haven't built a pmpy to deploy with ruby 2.7 yet.  We haven't done that because there are some changes to pmpy on a server called chard that provide some alternate logging that haven't made it into this code base and in a rubygem release yet.  It would be least risky to have the option to build the next release against either 2.6 OR 2.7.

I probably made this change too big.  We noticed when we tried to merge the logging changes that the CI was complaining about a number of things.  This PR fixes that as well.  I can create a PR for just the CI fixes and then the Ruby version changes can go to 2.7 and be in a subsequent release rather than hedging or living in the past.

.rubocop.yml
pgwillia: Funny story.  I tried to make this 2.7 thinking that linting would be independent, but one of the nags was that it should match the required_ruby_version in pushmi_pullyu.gemspec.

pushmi_pullyu.gemspec
pgwillia: That's ready to review: https://github.com/ualbertalib/pushmi_pullyu/pull/253

.github/workflows/ruby.yml
lagoan: I ran into this error too, glad to have read you had this issue earlier 😄 

I see that you include the ``Gemfile.lock`` in this PR, will it be a problem making the dependency more rigid as you mentioned?

pgwillia: I changed my mind on this after discussion on Dev Edmonton Slack.  Shane contributed to that discussion.  Even the thinking in bundler has changed about this and committing Gemfile.lock is best practice.  If you need more flexible dependencies there are other ways.  For example in the Gemfile or running test in a matrix against a variety of Gemfile.lock.

lagoan: Sorry I missed that explanation earlier 😄

lagoan (approved):
Buen trabajo!

Issue #242, opened 2021-12-15
Clean up after failed deposit in swift 


pgwillia:
We should finish this TODO and use this space to clean up if the deposit fails and re-enqueue failures so that we don't miss preserving of our data. 
https://github.com/ualbertalib/pushmi_pullyu/blob/2e9610bc577ad23e0a3b6fa41bd4653bafec6031/lib/pushmi_pullyu/cli.rb#L194-L207

The consequence for not cleaning up is filling the disk.

In a test batch we saw the application consistently failing on line 26:
https://github.com/ualbertalib/pushmi_pullyu/blob/2e9610bc577ad23e0a3b6fa41bd4653bafec6031/lib/pushmi_pullyu/swift_depositer.rb#L21-L59

We have also observed failures when authentication tokens expire or there is a connection issue with Swift.

pgwillia:
Is it reaching this?
https://github.com/ualbertalib/pushmi_pullyu/blob/2e9610bc577ad23e0a3b6fa41bd4653bafec6031/lib/pushmi_pullyu/aip.rb#L21-L22

pgwillia:
https://rollbar.com/ualbertalib/pushmi_pullyu/items/78/

PR #257, opened 2022-11-01
Clean up after failed swift deposit

lagoan:
## Context

When an AIP failed to be deposited, the files in the working directory were not deleted; this caused the filesystem to fill up.

We are now ensuring the files are deleted all the time.

Related to #242 

## What's New

We are catching the errors from the AIP creation and ensuring the files are deleted afterwards.


lib/pushmi_pullyu/aip.rb
```diff
@@ -12,13 +12,16 @@ def create(entity)
...
 
-    PushmiPullyu::AIP::Downloader.new(entity, aip_directory).run
-    PushmiPullyu::AIP::Creator.new(entity[:uuid], aip_directory, aip_filename).run
-
-    yield aip_filename, aip_directory
-
-    FileUtils.rm_rf(aip_filename) if File.exist?(aip_filename)
-    FileUtils.rm_rf(aip_directory) if File.exist?(aip_directory)
+      yield aip_filename, aip_directory
+    # Leave expection handling to logic at a higher level
```
pgwillia, line 20: I'm not sure what this comment is referring to?  Exceptions will be handled by the ensure block.  Which logic at what level?

pgwillia (changes requested):
Very good!  I think this is really clear and makes good use of some of the features of Ruby.  The test is clear and illustrative.

There's a comment that leaves me wondering if there was something missing from the PR.  An answer to that question would be all I need to approve this change.

For rubocop corrections I've been trying to include [why the changes are made](https://github.com/ualbertalib/pushmi_pullyu/pull/250) as part of the commit message.  What do you think of that approach?

lib/pushmi_pullyu/aip.rb
lagoan: I am using a ``begin`` block to add an ``ensure`` block in order to remove the files if something goes wrong. I am purposely not adding a ``rescue`` block that only re-raises the exception to be handled elsewhere. 

In our current version this is watched here:

https://github.com/ualbertalib/pushmi_pullyu/blob/c24aba37da054fe542c9293b96e5aa9930b557fb/lib/pushmi_pullyu/cli.rb#L203-L209

That piece of code is where the next work will be added to re-add the ids of the items that were not processed correctly.

I should make that comment more descriptive.

lagoan: To continue with our conversation:

The ensure part of the creation method is a sort of garbage collection for temporary files, there is no rescue block there, so essentially i am just re-raising the exception.

if an exception occurs it goes up to the ``run_preservation_cycle`` method in ``cli``

https://github.com/ualbertalib/pushmi_pullyu/blob/25f2ed564ab3ed1d75cca3ec7532aec36820c904/lib/pushmi_pullyu/cli.rb#L184-L210

The following piece of code is where the exception is caught

https://github.com/ualbertalib/pushmi_pullyu/blob/25f2ed564ab3ed1d75cca3ec7532aec36820c904/lib/pushmi_pullyu/cli.rb#L193-L209

it is in that rescue ``Exception => e`` block where I will re-add to redis the id of the item that failed to be preserved

It may be clearer to re-raise the exception. The only benefit of not explicitly adding that part is to have fewer lines. If fewer lines come at the cost of legibility, I will add that rescue block. @pgwillia, do you think this would be a better approach?

An example of how the exception is handled can be more apparent in its test

https://github.com/ualbertalib/pushmi_pullyu/blob/25f2ed564ab3ed1d75cca3ec7532aec36820c904/spec/pushmi_pullyu/aip_spec.rb#L58-L85

Specifically on this piece of code:

https://github.com/ualbertalib/pushmi_pullyu/blob/25f2ed564ab3ed1d75cca3ec7532aec36820c904/spec/pushmi_pullyu/aip_spec.rb#L71-L84

Here I pass a block that simply raises a ``StandardError`` and we check the files were removed by the ``ensure`` block.

Happy to hear your thoughts on this.

lib/pushmi_pullyu/aip.rb
```diff
@@ -12,13 +12,17 @@ def create(entity)
...
-    PushmiPullyu::AIP::Creator.new(entity[:uuid], aip_directory, aip_filename).run
-
-    yield aip_filename, aip_directory
-
-    FileUtils.rm_rf(aip_filename)
-    FileUtils.rm_rf(aip_directory)
+      yield aip_filename, aip_directory
+    # Here we will ensure the files are removed even if an exception comes up.
+    # We will leave the exception handling when we actually create an AIP using
+    # this method.
```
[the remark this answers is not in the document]
lagoan: Much clearer, thanks!

pgwillia (approved):

Issue #232, opened 2021-11-17
redis queue is drained by PMPY even though items in the queue are not preserved with success


henryzhang87:
step 1: push items into redis database (on era-app-stg-1.library.ualberta.ca)

```
RAILS_ENV=staging rails jupiter:preserve_all_items_and_theses[2021-01-28]
Adding Items and Theses on or after 2021-01-28 to preservation queue...
2598 Items and 1246 Theses have been added to preservation queue!
```

step 2: verify redis queue ( showing the last few items on test redis server: reading.library.ualberta.ca)

```
[root@reading ~]#  redis-cli -a PASSWORD -n 2 zrange staging:pmpy_queue 0 50000

.....
3836) "{\"uuid\":\"ff2a9872-c25f-4bd5-8469-7120ae3342e5\",\"type\":\"theses\"}"
3837) "{\"uuid\":\"ff2f4e64-2b02-450a-878b-726d7c0d0abc\",\"type\":\"theses\"}"
3838) "{\"uuid\":\"ff59b26a-b324-4154-8cc3-2be1d23b47a5\",\"type\":\"theses\"}"
3839) "{\"uuid\":\"ff6292e9-ce86-4b5f-8e43-897a5bccab25\",\"type\":\"theses\"}"
3840) "{\"uuid\":\"ff632c89-e3b0-4fc3-bd47-29452a412388\",\"type\":\"theses\"}"
3841) "{\"uuid\":\"ff7ef4e8-79e5-405d-8ed7-bb3d3fcc9d97\",\"type\":\"theses\"}"
3842) "{\"uuid\":\"ff99aecb-4f3c-4dd7-817c-5b78093b605a\",\"type\":\"theses\"}"
3843) "{\"uuid\":\"ffa56755-088c-42fe-b984-04810dae0d01\",\"type\":\"theses\"}"
3844) "{\"uuid\":\"ffd79c4a-f7fc-430c-8243-4d0a52167ac0\",\"type\":\"theses\"}"

```

step 3: start PMPY service on middlesbrough (test server) with wrong configuration on purpose

After the service was started, two things happened.

The first thing is the draining of the redis "staging:pmpy_queue". It became empty

```
redis-cli -a PASSWORD -n 2 zrange staging:pmpy_queue 0 50000
(empty list or set)

```

The second thing is the logging of pushmi_pullyu, for each item, it has the following messages:

```
[...]

PR #269, opened 2022-11-18
Fix/re add item id after failed deposit

lagoan:
## Context

Entity uuid, and type information is consumed from the redis queue, and it is not re-added on failed deposit, loosing track of the entity for a future deposit.

Related to #232 

## What's New

Added the entity information back to the redis queue on a StandardError exception and cleaned up exception handling on the ``run_preservation_cycle`` method.

### DIFF
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -183,8 +183,8 @@ class PushmiPullyu::CLI
 
   def run_preservation_cycle
     entity_json = queue.wait_next_item
-    # jupiter is submitting the entries to reddis in a hash format using fat arrows. We need to change them to clons in
-    # order to parse them corretrly from json
+    # jupiter is submitting the entries to reddis in a hash format using fat arrows. We need to change them to colons in
+    # order to parse them correctly from json
     entity = JSON.parse(entity_json.gsub('=>', ':'), { symbolize_names: true })
     return unless entity[:type].present? && entity[:uuid].present?
 
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -198,14 +198,21 @@ class PushmiPullyu::CLI
         # Log successful preservation event to the log files
         PushmiPullyu::Logging.log_preservation_event(deposited_file, aip_directory)
       end
-      # rubocop:disable Lint/RescueException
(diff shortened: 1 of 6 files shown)


pgwillia (approved):
Looks good.  Liked that you thought through the exceptions from most specific (`EntityInvalid`) to least specific (`Exception`) and only re-enqueue the application's exceptions.

Some comments about the json parsing issue you found.  I suspect that your fix will continue to work if we fix it on the jupiter side or if Ruby 3.0 changes behaviour.

lib/pushmi_pullyu/cli.rb
```diff
@@ -182,11 +182,10 @@ def rotate_logs
...
 
   def run_preservation_cycle
-    entity_json = JSON.parse(queue.wait_next_item)
-    entity = {
-      type: entity_json['type'],
-      uuid: entity_json['uuid']
-    }
+    entity_json = queue.wait_next_item
+    # jupiter is submitting the entries to reddis in a hash format using fat arrows. We need to change them to clons in
+    # order to parse them corretrly from json
```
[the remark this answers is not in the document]
lagoan: Ideally we should fix this on both sides to make sure we are getting propper json. I thought of leaving it as is to deploy it quickly an fix the issue on both sides. I will add a ticket on both projects so we don't forget to make this change.

spec/pushmi_pullyu/cli_spec.rb
```diff
@@ -272,6 +272,30 @@
 
         expect(old_options).to eq new_options
       end
+
+      it 'makes sure an entities information is readded to reddis when deposit fails' do
+        cli.parse(['-C', 'spec/fixtures/config_wrong_swift.yml'])
+        redis = Redis.new
+        redis.zadd(PushmiPullyu.options[:queue_name], 10,
+                   '{"uuid": "123e4567-e89b-12d3-a456-426614174000", "type": "items"}')
```
pgwillia, line 280: Is this the expected form, without the hash rockets?

pgwillia (approved):
:+1:

spec/pushmi_pullyu/cli_spec.rb
lagoan: my thinking is that we should be expect it to be standard json, if for whatever reason we change to another type of platform.

We need to change this in both jupiter and PMPY to get the right behaviour.

PR #270, opened 2022-11-21
Update CHANGELOG with recent dependabot updates

lagoan:
## Context

Updates to the changelog with information about recent gem updates from dependabot.


pgwillia (approved):
:+1:

PR #273, opened 2022-11-22
Bump CHANGELOG and VERSION for 2.0.4 release

lagoan:
## Context

We need cut a release that makes pushmi_pullyu more robust when a deposit fails in part of the workflow.

## What's New

Bump version and CHANGELOG to 2.0.4

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    pushmi_pullyu (2.0.3)
+    pushmi_pullyu (2.0.4)
       activesupport (>= 5, < 8)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -161,7 +161,7 @@ GEM
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
     rexml (3.2.5)
-    rollbar (3.3.2)
+    rollbar (3.3.0)
```
pgwillia, line 164: Why the regresssion?

lagoan: This was an automatic change. Will add a note to review in next change.

PR #279, opened 2022-12-15
Add exception handling to preservation cycle

lagoan:
We are getting errors on the preservation cycle while waiting for the
next item in the redis queue. This small change will make sure we log
the problem so we can track it.


pgwillia (approved):
:+1: LGTM.  We're just adding some logging in here to see what errors are being raised while waiting for the next item and parsing the type and uuid.

Issue #255, opened 2022-08-09
Adding communities and collection metadata to AIPs


pgwillia:
Currently AIPs are not being created for collections. There is important information about collections that we should be capturing in the preservation metadata in one way or another. This discussion can pick up again following on conversations about our preservation path going forward.

Reviewed details on what is currently being collected in the database for communities and collections (database schema: https://github.com/ualbertalib/jupiter/blob/b6961f5a3b9fd20fd552469d5872eca425f7d6bf/db/schema.rb#L91-L123

samples: https://gist.github.com/pgwillia/eed7dd858e17a9a67f9d90cb1d703adb#file-atmire_collection_data-csv and https://gist.github.com/pgwillia/eed7dd858e17a9a67f9d90cb1d703adb#file-atmire_community_data-csv). 

Properties for inclusion in AIPs for collections:

rdf:type (i.e., is a pcdm:Collection)
pcdm:memberOf (uuid for community)
dc:title
dc:description
dc:accessRights
ual:restrictedCollection
info:createdDate
ual:recordCreatedInJupiter

Properties for inclusion in AIPs for communities:

rdf:type (is a pcdm:Object, is a ual:Community)
dc:title
dc:description
dc:accessRights
info:createdDate
ual:recordCreatedInJupiter

_Originally posted by @sfarnel in https://github.com/ualbertalib/pushmi_pullyu/issues/196#issuecomment-1209738425_

pgwillia:
We discussed a couple ways to do this
1. preserving collections and communities separately
2. adding collection and community metadata to the AIPs we preserve

We were leaning towards #2 but would have to enque all items in a community/collection any time it changed.

lagoan:
I am trying to figure out the best place to store that metadata for the communities and collections.

Looking at the Item model I see and item can belong to multiple community/collection pairs. Going with [option 2](https://github.com/ualbertalib/pushmi_pullyu/issues/255#issuecomment-1209819285) would mean that the information for all community/collection pairs would reside in the AIP.

Would the best place for that metadata be in the ``object_metadata.n3`` file? Here is an [example](https://github.com/ualbertalib/jupiter/blob/master/test/fixtures/files/n3/items/e2ec88e3-3266-4e95-8575-8b04fac2a679-base.n3) of the current contents of that file.

[...]

PR #291, opened 2023-01-19
Wip add community and collection to aips

lagoan:
## Context

Collection and community models were not being fetched for AIP creation and were not saved for preservation. The latest changes on jupiter will open up community and collection information. This change makes PMPY aware of this and will handle AIP creation.

Related to #255 

## What's New

Added logic to handle the creation of community and collection AIPs. These AIPs are very similar to the ones created for items and theses, with the exception that they will never include files.

### DIFF
--- a/lib/pushmi_pullyu/aip/downloader.rb
+++ b/lib/pushmi_pullyu/aip/downloader.rb
@@ -199,7 +199,7 @@ class PushmiPullyu::AIP::Downloader
       main_object_remote: object_uri,
       main_object_local: "#{aip_dirs[:metadata]}/object_metadata.n3",
       file_sets_remote: "#{object_uri}/filesets",
-      # This directory needs to be created before we can dowloand the file order information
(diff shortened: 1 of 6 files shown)


lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -193,7 +199,9 @@ def object_aip_paths
       main_object_remote: object_uri,
       main_object_local: "#{aip_dirs[:metadata]}/object_metadata.n3",
       file_sets_remote: "#{object_uri}/filesets",
-      file_sets_local: "#{aip_dirs[:files_metadata]}/file_order.xml",
+      # This directory needs to be created before we can dowloand the file order information
```
ConnorSheremeta, line 202: Small typo in this comment here

lagoan: Thanks!

ConnorSheremeta (approved):
LGTM

Issue #297, opened 2023-01-30
Re-ingestion attempts are happening too quickly flooding the logs


lagoan:
The new functionality that re-adds items that failed to be ingested floods the logs. This happens because the retries happen right after the failed attempt with only a 10-second delay.

We need to add a way to delay the attempts from PMPY to reingest the entities. We talked about doubling the time between retries. With a max wait time of a week.

jefferya:
I'm not sure where to add this comment. I think the approach good. In case it helps, I'll share what was done with the CWRC preservation workflow. Adding delays helped with the CWRC preservation however failures still happened. A choice was made to assume failures would happen and development went toward both an audit tool to detect failures and report item preservation status plus a mechanism to preserve individual failed items.

lagoan:
Thanks for sharing this @jefferya. Yes, we need to assume failures will continue to happen and we do need some sort of audit tool.  Would you be able to provide design documentation for the approach used in CWRC?

jefferya:
> Would you be able to provide design documentation for the approach used in CWRC?

The design of the current approach is at the following link: <https://github.com/ualbertalib/cwrc_preservation#reporting--auditing-cwrc_audit_reportrb>. The approach is simplistic. The audit report lists all IDs in CWRC and Swift with an indication  whether preservation was successful or absent (either due to a failure or due to the CWRC object being updated after the last preservation run) plus a quick check if the Swift object size makes sense. 

Here is a summary:
* a custom metadate field is added to the swift object at ingest time (not into the package but added to the swift object metadata) to indicate CWRC resource version
* Audit process:
  * API call to get a list of CWRC resources and their associated version
  * API call to get a list of Swift objects with the resource version from the Swift object metadata
  * compare the two lists aligning by ID and version (not very efficient because quickly done)
  * output a CSV report indicating
    * if matching ID and version then preservation was successful and up-to-date
    * if matching ID but outdated Swift version then either 
       * a preservation error
       * or a newly added/updated CWRC object since last preservation run
    * if only in CWRC then either 
      * a preservation error 
      * or a newly added CWRC object since the last preservation run
    * if only in Swift then this is CWRC deleted object (policy is to retain deleted within preservation, at present)
    * quick check of size (i.e., if size is below a minimum then likely the preservation package is an HTTP error as opposed to real content
[...]

PR #298, opened 2023-01-30
Add delay on entity re-ingestion on failed attempt

lagoan:
When an entity fails to be deposited, PMPY would only wait for the minimum age to retry to ingest the entity. This creates the problem where we don't let the problem be fixed and we flood the logs with information that is no longer useful.

This change doubles the time of each attempt to give space for the problems to be fixed. There are 3 configuration parameters added:

- **ingestion_prefix**: The prefix used for Redis keys storing the failed attempt count - Default value of  - Default value of `prod:pmpy_ingest_attempt:`
- **ingestion_attempts**: The max number of re-ingestion attempts - Default value of `15`
- **first_failed_wait**: The time to add at the beginning which will be double every retry - Default value of `10`

The default values will keep retrying for around a week and drop the entity for ingestion.

This is related to issue [#297](https://github.com/ualbertalib/pushmi_pullyu/issues/297).

### DIFF
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -148,6 +148,21 @@ class PushmiPullyu::CLI
         opts[:queue_name] = queue
       end
 
+      o.on('-i', '--ingestion_prefix PREFIX',
+           'Prefix for keys used in counting the number of failed ingestion attempts') do |prefix|
+        opts[:ingestion_prefix] = prefix
+      end
+
+      o.on('-x', '--ingestion_attempts NUMBER', Integer,
+           'Max number of attempts to try ingesting an entity') do |ingestion_attempts|
+        opts[:ingestion_attempts] = ingestion_attempts
+      end
+
+      o.on('-f', '--first_failed_wait NUMBER', Integer,
+           'Time in seconds to wait after first failed deposit. Time will double every failed attempt') do |failed_wait|
+        opts[:first_failed_wait] = failed_wait
+      end
+
       o.separator ''
       o.separator 'Common options:'
 
@@ -183,16 +198,10 @@ class PushmiPullyu::CLI
 
   def run_preservation_cycle
     begin
-      entity_json = queue.wait_next_item
-      return unless entity_json
-
-      # jupiter is submitting the entries to reddis in a hash format using fat arrows. We need to change them to colons
-      # in order to parse them correctly from json
-      entity = JSON.parse(entity_json.gsub('=>', ':'), { symbolize_names: true })
-      return unless entity[:type].present? && entity[:uuid].present?
+      entity = queue.wait_next_item
+      return unless entity && entity[:type].present? && entity[:uuid].present?
     rescue StandardError => e
-      Rollbar.error(e)
-      logger.error(e)
+      log_exception(e)
     end
 
     # add additional information about the error context to errors that occur while processing this item.
@@ -209,7 +218,11 @@ class PushmiPullyu::CLI
     # readding it to the queue as it will always fail
     rescue PushmiPullyu::AIP::EntityInvalid => e
     rescue StandardError => e
-      queue.add_entity_json(entity_json)
+      begin
+        queue.add_entity_in_timeframe(entity)
+      rescue MaxDepositAttemptsReached => e
+        log_exception(e)
+      end
 
     # rubocop:disable Lint/RescueException
     # Something other than a StandardError exception means something happened which we were not expecting!
@@ -218,8 +231,7 @@ class PushmiPullyu::CLI
       raise e
     # rubocop:enable Lint/RescueException
     ensure
-      Rollbar.error(e)
-      logger.error(e)
+      log_exception(e)
     end
   end
 
@@ -292,6 +304,11 @@ class PushmiPullyu::CLI
       Dir.chdir(pwd)
       start_server
     end
+  end
+
+  def log_exception(exception)
+    Rollbar.error(exception)
+    logger.error(exception)
   end
 
 end
--- a/lib/pushmi_pullyu/preservation_queue.rb
+++ b/lib/pushmi_pullyu/preservation_queue.rb
@@ -20,6 +20,7 @@ require 'connection_pool'
 class PushmiPullyu::PreservationQueue
 
   class ConnectionError < StandardError; end
+  class MaxDepositAttemptsReached < StandardError; end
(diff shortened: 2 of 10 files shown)


lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -68,10 +70,29 @@ def wait_next_item
...
+    end
+  end
+
+  def self.extra_wait_time(deposit_attempt)
+    extra = 0
+    deposit_attempt.times do |n|
+      extra += (2**n) * PushmiPullyu.options[:first_failed_wait]
     end
+    extra
   end
```
ConnorSheremeta, line 96: This doesn't appear to match the stated specifications of: "doubles the time of each attempt"

With a PushmiPullyu.options[:first_failed_wait] of 10 it appears that this is the added wait time of each attempt:
```
0
10
30
70
150
310
630
1270
2550
5110
```
If doubling added wait on each attempt starting with first_failed_wait is what we're trying to accomplish it looks like this would give that:
```
def self.extra_wait_time(deposit_attempt)
  return 0 if deposit_attempt == 0

  (2**(deposit_attempt - 1)) * PushmiPullyu.options[:first_failed_wait]
end
```
The numbers vary slightly so this is a small nit in reality

lib/pushmi_pullyu/cli.rb
```diff
@@ -148,6 +148,21 @@ def parse_options(argv)
...
+      o.on('-x', '--ingestion_attempts NUMBER', Integer,
+           'Max number of attempts to try ingesting an entity') do |ingestion_attempts|
+        opts[:ingestion_attempts] = ingestion_attempts
+      end
+
+      o.on('-f', '--first_failed_wait NUMBER', Integer,
+           'Time in seconds to wait after first failed deposit. Time will double every failed attempt') do |failed_wait|
+        opts[:first_failed_wait] = failed_wait
+      end
+
```
ConnorSheremeta, line 165: Probably worth adding these new options to the [README](https://github.com/ualbertalib/pushmi_pullyu#usage)

lib/pushmi_pullyu/preservation_queue.rb
lagoan: You are right!

This code should work:

```ruby
(2**deposit_attempt) * PushmiPullyu.options[:first_failed_wait]
```

Will make the changes

ConnorSheremeta (approved):
LGTM. I was curious if items that fail should be kept track of? Not sure if there's already a way to verify which items didn't succeed or if we even care to keep track of this.

lib/pushmi_pullyu/cli.rb
```diff
@@ -209,7 +218,11 @@ def run_preservation_cycle
     # readding it to the queue as it will always fail
     rescue PushmiPullyu::AIP::EntityInvalid => e
     rescue StandardError => e
-      queue.add_entity_json(entity_json)
+      begin
+        queue.add_entity_in_timeframe(entity)
+      rescue MaxDepositAttemptsReached => e
+        log_exception(e)
+      end
```
lagoan, line 225: This bit of code will log when an entity reaches the max number of attempts for ingestion.

lagoan: @ConnorSheremeta do you think this would be enough to keep track of the failed items? It would mean going through the logs though, not sure if this is the best approach. Always open to ideas!

ConnorSheremeta: Filtering through the logs could suffice although it may not be ideal, I suppose these failures could be logged to another file but that may not be worth doing as this information can be filtered as needed

spec/pushmi_pullyu/preservation_queue_spec.rb
```diff
@@ -8,20 +8,24 @@
 
     before do
       PushmiPullyu.server_running = true
-      redis.zadd 'test:pmpy_queue', 1, 'noid1'
-      redis.zadd 'test:pmpy_queue', 3, 'noid3'
-      redis.zadd 'test:pmpy_queue', 4, 'noid2'
```
pgwillia, line 13: Are `noid1`, `noid2`, and `noid3` still used or defined somewhere?

lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -68,12 +70,27 @@ def wait_next_item
...
+                        entity.slice(:uuid, :type).to_json
+      else
+        connection.del entity_attempts_key
+        raise MaxDepositAttemptsReached
+      end
     end
   end
 
+  def self.extra_wait_time(deposit_attempt)
+    (2**deposit_attempt) * PushmiPullyu.options[:first_failed_wait]
```
pgwillia, line 91: 👍

lib/pushmi_pullyu/cli.rb
```diff
@@ -183,16 +198,10 @@ def rotate_logs
 
   def run_preservation_cycle
     begin
-      entity_json = queue.wait_next_item
-      return unless entity_json
-
-      # jupiter is submitting the entries to reddis in a hash format using fat arrows. We need to change them to colons
-      # in order to parse them correctly from json
-      entity = JSON.parse(entity_json.gsub('=>', ':'), { symbolize_names: true })
```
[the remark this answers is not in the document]
lagoan: To my understanding, this was an old error where the items were being added to the queue in an incorrect manner that added the fat arrows. The new versions of jupiter and pmpy will be adding correct json.

lib/pushmi_pullyu/preservation_queue.rb
```diff
@@ -68,12 +70,27 @@ def wait_next_item
...
   end
 
-  def add_entity_json(entity_json)
+  def add_entity_in_timeframe(entity)
+    entity_attempts_key = "#{PushmiPullyu.options[:ingestion_prefix]}#{entity[:uuid]}"
+
     @redis.with do |connection|
-      connection.zadd @queue_name, Time.now.to_f, entity_json
+      # separate information for priority information and queue
+      deposit_attempt = connection.incr entity_attempts_key
```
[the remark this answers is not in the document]
lagoan: That is correct, the normal behaviour would continue

spec/pushmi_pullyu/preservation_queue_spec.rb
lagoan: These values where just used for this test.

PR #308, opened 2023-03-16
Solve communication problems with jupiter

lagoan:
## Context

Please describe any relevant motivation and context for these changes.

PMPY was not working in staging and all preservation attempts failed. PMPY could not find the full path to jupiter because it was using the URI request_uri which is the path after the localhost. If PMPY and jupiter are on different servers, this will not work.

Related to #297

## What's New

Changed use of method URI.request_uri to joining URI with new path.

### DIFF
--- a/lib/pushmi_pullyu/aip/downloader.rb
+++ b/lib/pushmi_pullyu/aip/downloader.rb
@@ -86,7 +86,7 @@ class PushmiPullyu::AIP::Downloader
     @uri = URI.parse(PushmiPullyu.options[:jupiter][:jupiter_url])
     @http = Net::HTTP.new(@uri.host, @uri.port)
     @http.use_ssl = true if @uri.instance_of? URI::HTTPS
-    request = Net::HTTP::Post.new(URI.join(@uri,'/auth/system'))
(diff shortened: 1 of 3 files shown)


lib/pushmi_pullyu/aip/downloader.rb
```diff
@@ -86,7 +86,7 @@ def authenticate_http_calls
     @uri = URI.parse(PushmiPullyu.options[:jupiter][:jupiter_url])
     @http = Net::HTTP.new(@uri.host, @uri.port)
     @http.use_ssl = true if @uri.instance_of? URI::HTTPS
-    request = Net::HTTP::Post.new("#{@uri.request_uri}auth/system")
+    request = Net::HTTP::Post.new(URI.join(@uri,'/auth/system'))
```
ConnorSheremeta, line 89: LGTM except for this robocop violation

ConnorSheremeta (approved):

lib/pushmi_pullyu/aip/downloader.rb
lagoan: Fixed!

PR #310, opened 2023-03-17
Prepare for 2.0.6 release

lagoan:
## Context

Prepare for 2.0.6 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and releases in PushmiPullyu adheres to [Semantic Versioning](https://semver.org
 
 ## [Unreleased]
 
+## [2.0.6] - 2023-03-17
+
 - Fix URI concatenation for jupiter's base url. [#309](https://github.com/ualbertalib/pushmi_pullyu/issues/309)
 
 ## [2.0.5] - 2023-02-17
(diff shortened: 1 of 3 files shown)


ConnorSheremeta (approved):
LGTM, noticed that bundler was downgraded from 2.3.19 to 2.3.12 however, was this intended?

lagoan:
> LGTM, noticed that bundler was downgraded from 2.3.19 to 2.3.12 however, was this intended?

Thanks for bringing that up! It was not intended on my part. I tried to revert it to 2.3.19 by creating a new gemset for rvm, and installing 2.3.19 specifically to install packages but could it did not change the Gemfile.lock entry. 

Do you have an idea of how to force version 2.3.19 without manually editing the file? Thanks for your feedback!

ConnorSheremeta:
> > LGTM, noticed that bundler was downgraded from 2.3.19 to 2.3.12 however, was this intended?
> 
> Thanks for bringing that up! It was not intended on my part. I tried to revert it to 2.3.19 by creating a new gemset for rvm, and installing 2.3.19 specifically to install packages but could it did not change the Gemfile.lock entry.
> 
> Do you have an idea of how to force version 2.3.19 without manually editing the file? Thanks for your feedback!

I think getting the correct bundler version through gem install followed by running bundle install would update gemfile.lock to the version that was installed. I'm on rbenv however so there may be some discrepancies

Issue #314, opened 2023-04-19
 ERROR -- : [Rollbar] Tried to send a report with no message, exception or extra data


henryzhang87:
As the title shows, rollbar tries to send an empty message, no dump of stack trace. These messages are not helpful. Wondering if it is a good idea to stop it. 

Below are some sample outputs related to the error message:

E, [2023-04-11T10:50:28.161736 #1913] ERROR -- : [Rollbar] Tried to send a report with no message, exception or extra data. E, [2023-04-11T11:00:10.441705 #1913] ERROR -- : [Rollbar] Tried to send a report with no message, exception or extra data. E, [2023-04-12T08:42:50.424814 #1913] ERROR -- : [Rollbar] Tried to send a report with no message, exception or extra data. 

In total, there are **74420** records in the current logging folder as shown below:

 zcat /var/pushmi_pullyu/log/pushmi_pullyu.log-*.gz | grep "Tried to send a report with no message" | wc -l 
**74420**

If it is desired to stop this, a check to see if the exception is nil, if so, then bypass it.

original source code in cli.rb:

```
  def log_exception(exception)
    Rollbar.error(exception)
    logger.error(exception)
  end
```

PR #341, opened 2023-06-22
Make sure the exception is in the scope

lagoan:
## Context

We had recurring nil exceptions being logged. This was due to a basic error in the code scope.

Related to #314 

## What's New

Moved log method to the correct scope.


jefferya (approved):
👍

Issue #349, opened 2023-07-05
Investigate using pushmi_pullyu with OLRC


pgwillia:
Investigate using pushmi_pullyu with OLRC

I think we will likely need to re-configure PMPY to point to OLRC this fall.  With that we might take several approaches - simple configuration of existing PMPY against a new swift container using the current API approach, or a more complex project of integrating Duracloud into our PMPY pipeline. - Kenton

First steps
1. Meet with Kenton for context and make this a ticket.
2. Ask for credentials that we'll need and the container we'll write to for this test.

Maybe the first step is seeing if our development environment will write to OLRC.

jefferya:
Here are the major differences I've experienced when converting the CWRC perseveration tooling to use OLRC Swift as a drop-in replacement for UAL Swift.

The main difference, OLRC requires an additional property, a domain as part of the authentication. As of 2023-08-15, I'm testing to see if further changes are required: <https://github.com/ualbertalib/cwrc_preservation/tree/jefferya/olrc_compatibility>

* OLRC OpenStack RC file env var `OS_USER_DOMAIN_NAME` to OpenStack Gem connection property `user_domain`
* OLRC OpenStack RC file env var `OS_PROJECT_DOMAIN_ID` to OpenStack Gem connection property `project_domain_id`
 
The ruby-openstack connection is defined here: <https://github.com/ruby-openstack/ruby-openstack/blob/master/lib/openstack/connection.rb> 

The aforementioned OpenStack RC file can be downloaded from the Horizon UI via the upper right-hand corner user account menu.

lagoan:
Thanks to @jefferya work the process of redirecting PMPY to OLRC has been straightforward. 

The following shell command works for me to test authentication:

```bash
swift --os-auth-url 'AUTH_URL' \
      --os-username 'USER_NAME' \
      --os-user-domain-name 'USER_DOMAIN_NAME' \
      --os-password  'PASSWORD' \
      --os-project-name 'PROJECT_NAME' \
      --os-project-domain-name 'PROJECT_DOMAIN_NAME' \
      --auth-version 3 \
      stat
```

The changes in the codebase to redirect pushing the items to OLRC Swift have been mirroring the configuration changes used to authenticate. 

I've successfully pushed synthetic data to a new container in OLRC.

My next steps include:

- [ ] Testing pushing real data from a database dump in my development environment
- [ ] Test with real data, including files, from our staging environment
[...]

PR #422, opened 2024-03-14
Add support for V3 authentication for Swift

lagoan:
This PR adds configuration options to create Swift depositer objects configured with V3 authentication.

This configuration will be used with the move to OLRC.

## Context

We are moving Swift destination from local to OLRC. This requires a different type of authentication. The rest of the behaviour stays the same.

Related to #349

## What's New

Adds configuration options and tests for authenticating on Openstack Swift servers with V3 authentication.

### DIFF
--- a/lib/pushmi_pullyu.rb
+++ b/lib/pushmi_pullyu.rb
@@ -39,8 +39,8 @@ module PushmiPullyu
       auth_url: 'http://localhost:8080/auth/v1.0',
       project_name: 'demo',
       project_domain_name: 'default',
-      container: 'ERA',
-      auth_version: 'v1',
+      container: 'era',
+      auth_version: 'v3',
       user_domain: 'default'
     },
     rollbar: {},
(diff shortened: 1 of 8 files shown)


pgwillia (dismissed):
LGTM :+1: 

Just a curiosity about what the appropriate default should be.

lib/pushmi_pullyu.rb
```diff
@@ -39,7 +39,9 @@ module PushmiPullyu
       auth_url: 'http://localhost:8080/auth/v1.0',
       project_name: 'demo',
       project_domain_name: 'default',
-      container: 'ERA'
+      container: 'ERA',
+      auth_version: 'v1',
```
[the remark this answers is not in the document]
lagoan: Having v3 as the default was my initial approach, but considered having the current version as to not mess with current behaviour in case we want to go back to it with the transition.

Having v3 as default makes sense now that more people took a look at it.

Will make the change.

spec/fixtures/config.yml
```diff
@@ -18,6 +18,8 @@ rollbar:
   proxy_port: '80'
 swift:
   auth_url: http://127.0.0.1:8080/auth/v1.0
+  auth_version: 'v1'
+  container: 'ERA'
```
pgwillia, line 22: Does this need to be lower case?

lagoan: Not for 'v1' authentication which is what we are currently using.

pgwillia (approved):
LGTM :+1:

PR #443, opened 2024-06-05
Improve log clarity

lagoan:
This change makes logs easier to understand by humans.

We start counting ingestion attempts at 1.

We will also format all events in the JSON log with the same structure.

### DIFF
--- a/lib/pushmi_pullyu/cli.rb
+++ b/lib/pushmi_pullyu/cli.rb
@@ -199,8 +199,9 @@ class PushmiPullyu::CLI
   def run_preservation_cycle
     begin
       entity = queue.wait_next_item
+      # We add + 1 to the get_entity_ingestion_attempt because humans like to start counting from 1
       PushmiPullyu::Logging.log_preservation_attempt(entity,
-                                                     queue.get_entity_ingestion_attempt(entity))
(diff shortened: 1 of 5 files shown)


lib/pushmi_pullyu/cli.rb
```diff
@@ -199,8 +199,9 @@ def rotate_logs
   def run_preservation_cycle
     begin
       entity = queue.wait_next_item
+      # We add + 1 to the get_entity_ingestion_attempt because humans like to start counting from 1
       PushmiPullyu::Logging.log_preservation_attempt(entity,
-                                                     queue.get_entity_ingestion_attempt(entity))
+                                                     queue.get_entity_ingestion_attempt(entity) + 1)
```
ConnorSheremeta, line 204: Small nit: I feel like humanising the counts inside of the log_preservation_x methods would be ideal. Not a show stopper though

ConnorSheremeta (approved):
Looks good to me! I left a small nit which may be somewhat subjective. The suggested change may also have other impacts. I think this is good to be merged as is unless that change is worthwhile.

ConnorSheremeta (approved):
LGTM!

PR #457, opened 2024-07-08
Fix problem with authentication token expiration

lagoan:
## Context

We've been having multiple authentication token errors due to invalidation. This PR Fixes this

Related to #315 

## What's New

- Set a parameter default value that makes the ruby-openstack gem retry to refresh the auth token.
- Add an http header to refresh the token on each call.
- Catch the exception when a token is invalid and recreate the connection.

### DIFF
--- a/lib/pushmi_pullyu/swift_depositer.rb
+++ b/lib/pushmi_pullyu/swift_depositer.rb
@@ -62,8 +62,7 @@ class PushmiPullyu::SwiftDepositer
       # and metadata as additional key/value string pairs in the hash
       headers = { 'etag' => checksum,
                   'content-type' => 'application/x-tar',
-                  'X-Auth-New-Token' => 'true'
-                }.merge(metadata)
+                  'X-Auth-New-Token' => 'true' }.merge(metadata)
       deposited_file = era_container.object(file_base_name)
       deposited_file.write(File.open(file_name), headers)
     else
@@ -71,8 +70,7 @@ class PushmiPullyu::SwiftDepositer
       headers = { etag: checksum,
                   content_type: 'application/x-tar',
                   metadata: metadata,
-                  'X-Auth-New-Token' => 'true'
-                 }
+                  'X-Auth-New-Token' => 'true' }
       deposited_file = era_container.create_object(file_base_name, headers, File.open(file_name))
     end
 
--- a/lib/pushmi_pullyu/swift_depositer.rb
+++ b/lib/pushmi_pullyu/swift_depositer.rb
@@ -2,10 +2,9 @@ require 'digest/md5'
 require 'openstack'
 class PushmiPullyu::SwiftDepositer
 
-  def initialize(connection, connection_wait_time: 10)
-    @connection_wait_time = connection_wait_time
+  def initialize(connection)
     # Generic authentication parameters
-    @swift_connection_parameters = {
+    swift_connection_parameters = {
       username: connection[:username],
       api_key: connection[:password],
       auth_url: connection[:auth_url],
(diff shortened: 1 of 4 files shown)


pgwillia (commented):
This feels like two or three loosely related changes without any additional tests.  I'm seeing if I can figure out a way to suggest some tests.

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -1,37 +1,46 @@
...
 
   def deposit_file(file_name, swift_container)
     file_base_name = File.basename(file_name, '.*')
-
     checksum = Digest::MD5.file(file_name).hexdigest
-
-    era_container = swift_connection.container(swift_container)
+    begin
+      era_container = @swift_connection.container(swift_container)
+    rescue OpenStack::Exception::ExpiredAuthToken => e
```
pgwillia, line 36: With `@retry_auth` true is this redundant?

Seems similar to https://github.com/ruby-openstack/ruby-openstack/blob/373e0c47ab23b67c695873839282866d0849989a/lib/openstack/connection.rb#L203-L207

lib/pushmi_pullyu/swift_depositer.rb
```diff
@@ -52,18 +61,26 @@ def deposit_file(file_name, swift_container)
...
+                  'X-Auth-New-Token' => 'true' }.merge(metadata)
       deposited_file = era_container.object(file_base_name)
       deposited_file.write(File.open(file_name), headers)
     else
       # for creating new: construct hash with symbols as keys, add metadata as a hash within the header hash
       headers = { etag: checksum,
                   content_type: 'application/x-tar',
-                  metadata: metadata }
+                  metadata: metadata,
+                  'X-Auth-New-Token' => 'true' }
```
pgwillia, line 73: https://stackoverflow.com/a/16525684 applies to swauth, and I can't find mention of it in https://github.com/openstack/swift/ or https://github.com/ruby-openstack/ruby-openstack.  Do you know what auth we're using with local or OLRC Swift?

pgwillia:
> This feels like two or three loosely related changes without any additional tests. I'm seeing if I can figure out a way to suggest some tests.

Here's part of a test for the retry_auth config
https://github.com/ualbertalib/pushmi_pullyu/commit/8a0c21cc0e13303ae15bd81e4d250de1b7b75d69

lib/pushmi_pullyu/swift_depositer.rb
lagoan: You are right. However, I could test that exception in our code, not the OpenStack gem, so I left our retry.

I am worried about the gem not being maintained. This us causing an error at the moment.

I am open to removing my code that catches the exception in order to have less things to maintain on our side. Let me know what you think.

lagoan:
> This feels like two or three loosely related changes without any additional tests. I'm seeing if I can figure out a way to suggest some tests.

I can add tests that fake the exception coming from the OpenStack gem. I've done it in the past.

lib/pushmi_pullyu/swift_depositer.rb
pgwillia: We're using keystone v1 for local swift and keystone v3 for OLRC

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -40,7 +40,7 @@
 
           expect(first_deposit.name).to eq(second_deposit.name)
           expect(first_deposit.container.name).to eq(second_deposit.container.name)
-        end.to change { swift_depositer.swift_connection.container('ERA').count.to_i }.by(1)
+        end.to change { swift_depositer.instance_variable_get('@swift_connection').container('ERA').count.to_i }.by(1)
```
pgwillia, line 43: I think it was easier to read before, but I agree that having an attribute accessor is unnecessary just for this purpose.

pgwillia (approved):
LGTM :+1:

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -59,9 +59,21 @@
...
         expect(swift_depositer).not_to be_nil
         # rubocop:disable Layout/LineLength
-        expect(swift_depositer.swift_connection.connection.authtoken).to eq('gAAAAABl8hYAouKZJLkt8NDmuA2NjA1zOasGOAX-b2MfKpjiM_kf8sZHe42ipcs6Vb-57-aATajbTg54wIwhNhl2HKRfz5_rKfSJ0PnBQNFCVd4bKrdC0pHzoJMn9hkAa2tjBkqppBcMayvfqz-Ppxn0USnHw0z9zLLKDxGbRZwyhDJDhGOcIZg')
+        expect(swift_depositer.instance_variable_get('@swift_connection').connection.authtoken).to eq('gAAAAABl8hYAouKZJLkt8NDmuA2NjA1zOasGOAX-b2MfKpjiM_kf8sZHe42ipcs6Vb-57-aATajbTg54wIwhNhl2HKRfz5_rKfSJ0PnBQNFCVd4bKrdC0pHzoJMn9hkAa2tjBkqppBcMayvfqz-Ppxn0USnHw0z9zLLKDxGbRZwyhDJDhGOcIZg')
         # rubocop:enable Layout/LineLength
       end
     end
+
+    it 'uses ruby-openstack gem behaviour to refresh authentication token' do
+      VCR.use_cassette('swift_new_deposit') do
```
pgwillia, line 68: For repeatability this might need to be a different tape than swift_new_deposit.  Also maybe we should setup the docker container to set the `token_life`?

spec/pushmi_pullyu/swift_depositer_spec.rb
```diff
@@ -59,9 +59,21 @@
...
+        swift_depositer = PushmiPullyu::SwiftDepositer.new(username: 'test:tester',
+                                                           password: 'testing',
+                                                           tenant: 'tester',
+                                                           auth_url: 'http://127.0.0.1:8080/auth/v1.0',
+                                                           retry_auth: true)
+
+        expect(swift_depositer.inspect).to include '@retry_auth=true' # retry_auth isn't exposed
```
pgwillia, line 75: Alternately, perhaps we include this test in the existing test above?

PR #492, opened 2024-11-27
Permit versions of activesupport that work with Ruby 2.7

pgwillia:
## Context
We've had troubles deploying 2.1.4 in staging.  We observed it installing newer versions of gems versus what was in our Gemfile.lock that don't work in Ruby 2.7.

## What's New
Adjust pushmi_pullyu.gemspec to avoid gems that only work in Ruby 3+ environments.


lagoan (approved):
LGTM!

PR #494, opened 2024-12-06
Pin version of securerandom that work with Ruby 2.7

pgwillia:
## Context
Staging is blocked by deployment issue. Recent minor releases of gem dependencies have move to require Ruby 3.x because 2.7 is eol. We pin major versions in pushmi_pullyu.gemspec

Related to #492 

## What's New
Pin securerandom to 0.3.2 (0.4.0 depends on Ruby 3)


jefferya (approved):
LGTM!

